PR #2, opened 2025-10-08
setup fstab and paths

helms-charity:
Test URLs:
- After: https://fstab--idfc--aemsites.aem.page/


roperev (approved):
Let's do it

PR #14, opened 2025-10-25
Initial desktop header/nav functionality and styling

### LINKED ISSUE #4: header nav
Create the page content and update the header block as needed.

POC page for reference: https://idfcbank-optimization--internal-aem-eds-poc-en--tekno-point.aem.live/idfcfirstbank/nav

we'll only do the "global" nav in this ticket (yellow)

Image

roperev:
Fix #4

Test URLs:
- After: https://4-header--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

@helms-charity, they have a "secondary nav" functionality that can vary per page (each CC page is different, and the home page is different). I authored this into the page in a section with a particular class applied (currently `mid-banner` but after updating _sections.json in main I'll change it to `top-nav`) and then parsing that into the top of main just under the primary nav. Is this a good method (in terms of best practice, code management, etc.) to do something like this that can vary per-page? Would love to discuss it with you before I finish this. There's still more styling and functionality work on this of course.

### DIFF
--- a/blocks/secondary-navbar/secondary-navbar.css
+++ b/blocks/secondary-navbar/secondary-navbar.css
@@ -18,8 +18,8 @@
 /* Inner container - matches header nav exactly */
 .tabs-pane-js {
   max-width: 1440px;
-  margin: 0 auto;
-  padding: 0 32px;
+  position: relative;
+  left: 70px;
   box-sizing: border-box;
 }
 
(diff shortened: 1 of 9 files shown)


helms-charity:
For that L2 nav, currently they seem to have 50 templates with this nav on it. 
I'm thinking we can make these "L2" navigations as fragment pages. In page properties, we would create a property that has a Fragment path picker but use the path to the homepage L2 nav (that you authored) as the default. If they erase that path, then no L2 nav div is added.

blocks/secondary-navbar/secondary-navbar.css
```diff
@@ -1,125 +1,297 @@
...
+  gap: 10px;
+  padding: 20px 24px;
+}
+
+.menu-cardList.MT15 {
+  margin-top: 0;
+}
+
+/* Gradient Cards */
+.grdiantCard {
```
helms-charity, line 134: Is there a reason you're not using '.gradient' for the class name here? I'll approve and merge but maybe when we go to update the hexadecimal colors to css variable names we can fix the ESL spelling too.

helms-charity (approved):

PR #36, opened 2025-11-01
[UE] Add 1 accordion item with block for easier authoring

### LINKED ISSUE #32: Accordion block fix
Currently the UE doesn't show the accordion-items in the content tree, although you can add them.
There is an FAQ Accordion sample on https://idfcbank-optimization--internal-aem-eds-poc-en--tekno-point.aem.live/idfcfirstbank/ but you can't currently add children.

helms-charity:
I was able to see this working in the UE using ?ref. 
The first accordion item will be added by default when the block is added, with "Lorem Ipsum". This makes authoring a little faster/easier.
image

Subsequent accordion items will not have Lorem Ipsum.

Fix #32 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/accordion
- After: https://32-accordion--idfc--aemsites.aem.page/library/accordion

### DIFF
--- a/blocks/accordion/_accordion.json
+++ b/blocks/accordion/_accordion.json
@@ -10,7 +10,14 @@
             "template": {
               "name": "Accordion",
               "model": "accordion",
-              "filter": "accordion"
+              "filter": "accordion",
+              "item1": {
+                "sling:resourceType": "blocks/accordion/accordion-item",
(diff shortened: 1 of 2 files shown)


roperev (approved):
Good idea, I like that it improves the authoring experience.

PR #46, opened 2025-11-03
Initial banner UE json using composite multi-field for CTAs

### LINKED ISSUE #21: [UE] banner JSON properties
Each item needs these properties:
Image

roperev:
Be advised that multi-field and composite multi-field are new features, technically in early access, so they may have breaking changes. See https://www.aem.live/developer/component-model-definitions#multi-fields-and-composite-multi-fields. 

Fix #21

Test URLs:
- After: https://21-bannerUE--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/banner/_banner.json
+++ b/blocks/banner/_banner.json
@@ -1,0 +1,106 @@
+{
+    "definitions": [
+      {
+        "title": "Banner",
+        "id": "banner",
+        "plugins": {
+          "xwalk": {
+            "page": {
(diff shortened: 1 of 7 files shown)


helms-charity:
Per https://cq-dev.slack.com/archives/C06UDHGTV17/p1762195089404709, apparently multifields are not supported without a "sideport" so we'll see what happens.

helms-charity (approved):

Issue #22, opened 2025-10-30
[block] sticky credit card CTA CSS + JS


helms-charity:
https://www.idfcfirstbank.com/credit-card/rupay-credit-card
This is a sticky CTA at the bottom of the page with a translucent background.

<img width="614" height="546" alt="Image" src="https://github.com/user-attachments/assets/c21275b1-73e3-4a92-803a-d9027014b984" />

<img width="1248" height="135" alt="Image" src="https://github.com/user-attachments/assets/88ce19de-52e7-458a-840e-d7c5849b0cef" />

<img width="497" height="102" alt="Image" src="https://github.com/user-attachments/assets/16d5d122-6240-4e34-8969-10323257bcb9" />

PR #52, opened 2025-11-05
Add Sticky CTA block with mobile-first styles, JSON UE config, and js semantic tagging

roperev:
Fix #<gh-issue-id>

This one was primarily a Cursor one-shot prompt, I'll drop it here for future learnings and copy-pastability for popping out net new blocks quickly:

<pre>
create a new set of base files (js, css, json) for the brand new sticky-cta component in the blocks folder. Look at the format of other blocks. I want the _sticky-cta.json file to be properly instrumented for Adobe AEM Universal Editor, and I'd like the fields of the model to allow for the following 5 fields:

1. CTA URL (reference) 
2. CTA Text (text)
3. CTA id (text)
4. CTA Mobile Text (text)
5. CTA Subtitle (richtext)
</pre>

image

Will need a bit more styling but that will be handled in #22 once I can drop a component into my test page. 

Test URLs:
- After: https://23-stickyctaUE--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/sticky-cta/_sticky-cta.json
+++ b/blocks/sticky-cta/_sticky-cta.json
@@ -1,0 +1,60 @@
+{
+  "definitions": [
+    {
+      "title": "Sticky CTA",
+      "id": "sticky-cta",
+      "plugins": {
+        "xwalk": {
+          "page": {
+            "resourceType": "core/franklin/components/block/v1/block",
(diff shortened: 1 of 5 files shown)


helms-charity (approved):
Seems good from what I can see.

Issue #15, opened 2025-10-27
Secondary navbar block


helms-charity:
This is the sticky/anchor navbar as seen on this page:
https://idfcbank-optimization--internal-aem-eds-poc-en--tekno-point.aem.live/idfcfirstbank/

Issue #16, opened 2025-10-27
3rd level nav "nav cards" block


helms-charity:
See them in the dropdowns of the 3rd level of the header nav

<img width="1964" height="1004" alt="Image" src="https://github.com/user-attachments/assets/b4ebd497-5bbf-48b3-9470-d01ebebf096a" />

roperev:
In progress: https://experience.adobe.com/#/@sitesinternal/aem/editor/canvas/author-idfc-stage-65.adobecqms.net/content/idfc-edge/framework/ccnav.html?ref=16-categorynav

roperev:
<img width="1898" height="1198" alt="Image" src="https://github.com/user-attachments/assets/98f66f35-c441-4a31-b50f-1549b2607f8a" />

PR #57, opened 2025-11-06
nav item component

bunting-adbe:

Fix #15 

Test URLs:
- After: https://15-anchor-nav--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/anchor-nav/anchor-nav.js
+++ b/blocks/anchor-nav/anchor-nav.js
@@ -7,215 +7,198 @@
  * Build a card from a list item
  */
 function buildCard(item) {
-    const card = document.createElement('div');
-    card.classList.add('grdiantCard', 'grdP1');
-  
-    const link = item.querySelector('a');
-    if (!link) {
-      // Handle items without links (plain text)
-      const cardText = document.createElement('div');
-      cardText.classList.add('title');
-      cardText.textContent = item.textContent.trim();
-      card.appendChild(cardText);
-      return card;
-    }
-  
-    const cardLink = document.createElement('a');
-    cardLink.href = link.getAttribute('href') || '#';
-    cardLink.setAttribute('data-gtm-desk-l3cards-click', '');
-  
-    // Tags container (hidden by default for this structure)
-    const tagsContainer = document.createElement('div');
-    tagsContainer.classList.add('tags');
-    tagsContainer.style.display = 'none';
-    cardLink.appendChild(tagsContainer);
-  
-    // Card title
-    const title = document.createElement('div');
-    title.classList.add('title');
-    title.textContent = link.textContent.trim();
-  
-    const iconSpan = document.createElement('span');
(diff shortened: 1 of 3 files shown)


blocks/anchor-nav/anchor-nav.js
```diff
@@ -7,215 +7,198 @@
...
-  
-    const iconSpan = document.createElement('span');
-    iconSpan.classList.add('icon-Right');
-    title.appendChild(iconSpan);
-  
-    cardLink.appendChild(title);
-    card.appendChild(cardLink);
-  
+  const card = document.createElement('div');
+  card.classList.add('grdiantCard', 'grdP1');
```
roperev, line 11: We should fix these typos `grdiant` for `gradient` when we see them pop up, these came from TeknoPoint's POC mostly.

blocks/anchor-nav/anchor-nav.js
```diff
@@ -7,215 +7,198 @@
...
+  const iconSpan = document.createElement('span');
+  iconSpan.classList.add('icon-Right');
+  title.appendChild(iconSpan);
+
+  cardLink.appendChild(title);
+  card.appendChild(cardLink);
+
+  return card;
+}
+
```
roperev, line 47: I think this card and tag stuff is only going to be needed for the 3rd-level nav dropdown thing in #16. You probably can delete it from here to simplify.

blocks/anchor-nav/anchor-nav.js
```diff
@@ -7,215 +7,198 @@
...
+      // Add element to current group
+      currentGroup.elements.push(child);
     }
-  
-    return navItems;
+  });
+
+  // Add last group
+  if (currentGroup) {
+    navItems.push(currentGroup);
```
roperev, line 118: This is my own hacky code to read through the section on the page with `mid-banner` class applied to it and then pull out H1s into top-level parents and <li> items into children...you won't need any of this code either as your anchor-nav-item just directly has editable fields for this with no parsing/rewriting.

blocks/anchor-nav/anchor-nav.js
```diff
@@ -7,215 +7,198 @@
...
+    // No mid-banner section found, hide the block
+    block.style.display = 'none';
+    return;
   }
-  
-    // Optionally hide the mid-banner section since we've extracted its content
-    // midBannerSection.style.display = 'none';
+
+  // Parse content from mid-banner section
+  const navItems = parseMidBannerContent(midBannerSection);
```
roperev, line 190: This can probably be dropped too after removing lines 86-116 above as no extra parsing of in-page content needs to happen.

roperev (approved):
Some clean up and removal of extra code not needed (since you simplified it with normal UE component fields instead of trying to parse out an in-page section of H1 and li items) could be good, but I know this is still being worked on for styling, etc. This is good enough to merge to main to get it in UE.

PR #78, opened 2025-11-15
Fix animation for steps block

### LINKED ISSUE #37: [UE] Pay using UPI - animated "UPI steps"
This block looks similar to the UPI Steps block, except the icon items (and the lines connecting) appear one by one and animated. Maybe this can be the same block?

Image

roperev:
Added full support for section-level background color and image choice (could add gradient support if needed). Fixed all navigation, created all styling specific to "Pay with UPI" steps. 

Animation duration toggle in UE properties defines the duration of the entire set of animation, plus an initial 1 second delay when the section first scrolls into view. Currently set at 2-seconds in the authored page component, so total duration is 1s delay + 2s animation, etc. 

Fix #37 

Test URLs:
- After: https://37-animatedSteps--idfc--aemsites.aem.page/credit-card/rupay-credit-card

image

### DIFF
--- a/blocks/steps/_steps.json
+++ b/blocks/steps/_steps.json
@@ -55,17 +55,17 @@
         },
         {
           "component": "select",
-          "name": "animation",
+          "name": "classes",
           "label": "Animation",
-          "valueType": "boolean",
+          "valueType": "string",
           "options": [
             {
-              "name": "Animated",
-              "value": "true"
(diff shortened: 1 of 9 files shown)


bunting-adbe (approved):
looks ok to me

helms-charity (approved):

PR #87, opened 2025-11-18
25 card swiper feature UE changes

roperev:
Need to push these .json files into UE to be able to view and edit fields in UE. Also bringing in the SwiperJS js and css files to the library, along with their associated linting ignore rules. Can't finish building the JS logic very easily until I can get the UE property values into the DOM. 

Fix (partially 25 but not completely)

Test URLs:
- After: https://25-cardSwiperFeature--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 helix-importer-ui
 tools/*
 blocks/swipper/swipper-bundle.min.js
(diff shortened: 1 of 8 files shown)


bunting-adbe (approved):
looks good

PR #91, opened 2025-11-19
Add embed block functionality with JSON configuration, added autobloc…

### LINKED ISSUE #89: autoblock a fragment from a link
See https://github.com/helms-charity/mrvcopy/blob/main/scripts/scripts.js#L403-L415

Where if in a text editor, it is a linked address to a page under /fragments/, then it will be replaced with a Fragment block and will render whatever page content is in that fragment.

additional reference: See https://github.com/da-pilot/sling/blob/main/eds/scripts/utils.js#L147-L168

zhummerz:
…king functions to scripts.js

Fix #89 Test page has an embedded youtube link at the top. Also added a fragment FQDN link along with a local link to show that the content is embedded for both. 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/embed
- After: https://89-fragmentembed--idfc--aemsites.aem.page/library/embed

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -502,7 +502,7 @@ export function buildFragmentBlocks(main) {
  * @param {Element} main The container element
  */
 export function buildEmbedBlocks(main) {
-  const embedPlatforms = /youtu|vimeo|twitter\.com/;
+  const embedPlatforms = /youtu\.be|youtu|vimeo|twitter\.com/;
   main.querySelectorAll('a[href]').forEach((a) => {
     if (embedPlatforms.test(a.href) && linkTextIncludesHref(a)) {
       const embedBlock = buildBlock('embed', a.cloneNode(true));
@@ -739,6 +739,8 @@ export function decorateMain(main) {
   decorateSections(main);
   applySectionBackgroundColors(main);
   decorateBlocks(main);
+  buildFragmentBlocks(main);
+  buildEmbedBlocks(main);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


scripts/scripts.js
```diff
@@ -461,6 +463,55 @@ async function loadCategoryNavFragment(main) {
...
+      if (parent.tagName === 'P' && parent.querySelector('.block')) {
+        const div = document.createElement('div');
+        div.className = parent.className;
+        while (parent.firstChild) div.appendChild(parent.firstChild);
+        parent.replaceWith(div);
+      }
+    }
+  });
+}
+
```
roperev: We actually already have this functionality provided for with the `loadAutoBlock` function here: https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L564-L572. You can remove this function and the call on line 742 (`buildFragmentBlocks(main)`) as well.

scripts/scripts.js
```diff
@@ -688,6 +739,8 @@ export function decorateMain(main) {
   decorateSections(main);
   applySectionBackgroundColors(main);
   decorateBlocks(main);
+  buildFragmentBlocks(main);
```
roperev: You can remove this function as we already do this with auto blocking in this project.

roperev (changes requested):
We have the page fragment loading functionality with the auto blocking feature, so we don't need an explicitly defined `buildFragmentBlocks` function.

scripts/scripts.js
zhummerz: Hey Evan, I've removed the code mentioned above and am ready for another peer review.

helms-charity (approved):
Looks good

roperev (approved):
Looks great now.

PR #92, opened 2025-11-19
25 card swiper feature

### LINKED ISSUE #25: [block] Swiper cards refactor
I can see in the code there is a "swipper-card" block, and looking at the component-models.json, there exists a "swipper-card-item". However, there is no component-definition for either of these, so they don't work when added to a page.

Let's call the block "Swiper cards", and each item would be a "Swiper card".
Add a _swiper-cards.json for the UE definition/model/filters.

roperev:
Added handling specifically for testimonial-card variant, switched to minified versions at ~50% of the size of the non-minified js/css swiperJS libraries. Cleaned up UE display of extra "cards" properties and values. Updated logic to build section-style dataset information (like `data-swipable` and `data-starting-card`) to the actual div in the DOM for consistency. This function can be externalized to `scripts.js` in the future if we need it anywhere outside of the `cards` block later.

Fix #25

Test URLs:
- After: https://25-cardSwiperFeature--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Normal behavior puts the "active" card as the leftmost in whatever view (3, 2 or 1 card wide based on page width):
image

Testimonial-card gets the `centeredCards` swiperJS option, and the "active" card is then centered, so the first card in the deck starts in the center "slot" as intended and the center slot CSS class can be applied for the purple bg, etc.:
image


bunting-adbe (approved):
looks good to me

Issue #93, opened 2025-11-19
Replace "Inter" fixed fonts with updated variable width font


roperev:
In working on designing and styling various blocks, we noticed that we seem to have hard-coded font-weight versions of the "Inter" font used across the IDFC site. See https://github.com/aemsites/idfc/blob/main/styles/fonts.css, and pay attention to the `src: url` value for each font-face defined. This reduces our styling flexibility as we only have 3 total weights and have to switch using font-family which is non-intuitive and can lead to frustrating results. 

<img width="1406" height="763" alt="Image" src="https://github.com/user-attachments/assets/28132c92-cb2c-4355-9a40-40318685ddb2" />

We should replace these hardcoded variants with the freely available `InterVariable.woff2` file available from https://rsms.me/inter/ (download and extract from zip in the "web" folder). We will need to update the `fonts.css` to reference the new file name. We may also want to remove the `inter-semiBold` (which was recently added to map to the 500 weight) font-face definition. We should also update the CSS root variables that reference them at https://github.com/aemsites/idfc/blob/main/styles/styles.css#L55-L58. 

This change may break some styling on newly created blocks such as `cards`, `steps`, `faq-accordion`, `anchor-nav`, or `category-nav`, so those should be given a close before/after screenshot consideration and tweaked to update font properties as needed (or create additional issue tickets) if there are any changes introduced.

PR #94, opened 2025-11-20
Add 'All About' card variant and updated styles

### LINKED ISSUE #24: Swiper cards "all-about" variant
This ticket depends on the "Swiper cards refactor" ticket being done first.

Current use of 'swipper card':
https://idfcbank-optimization--internal-aem-eds-poc-en--tekno-point.aem.live/idfcfirstbank/

Image

We need an "all-about" variant (they are actually calling this "Rupay CC All About" in AEM).

Image

Image

roperev:
Added UE variant "All About" to choose from when adding a card block. 

Fix #24

Test URLs:
- After: https://24-allaboutcard--idfc--aemsites.aem.page/credit-card/rupay-credit-card

image

image

Two small bugs, our H2 font can't currently reach any heavier optical weight due to #93. Also, the images we download from their site aren't entirely transparent outside of the "character", they have a very light purple overlay (maybe 10% opacity?). Not sure where that's coming from but we may need to check that somewhere in the avif -> dynamic media -> png or other chain of download we're not introducing some sort of issue. I'll create a new bug ticket for that.


bunting-adbe (approved):
approved

PR #102, opened 2025-11-20
fix naming snafu

### LINKED ISSUE #30: [UE] Tabs-upi-link block
Start with a copy of the Tabs block, but we need a new one.

https://www.idfcfirstbank.com/credit-card/rupay-credit-card

Image

There is mobile vs. desktop text options for each tab. 

These are the properties of each block item:

Image

This page also has something very similar
https://www.idfcfirst.bank.in/credit-card/secured-rupay-credit-card

bunting-adbe:
fix missing model names

Fix #30 

Test URLs:
- After: https://<branch>--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/tabs-upi-link/_tabs-upi-link.json
+++ b/blocks/tabs-upi-link/_tabs-upi-link.json
@@ -36,7 +36,7 @@
     ],
     "models": [
       {
-        "id": "tabs",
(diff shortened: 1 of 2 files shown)


roperev (approved):
Good catch on the model name mismatch, I should have seen that on your prior PR!

PR #105, opened 2025-11-21
Add mobile and tablet hero block styling and logic

### LINKED ISSUE #97: [block] Hero CSS + JS
Customer hero block has a small amount of animation, the "card stack" image slides in from the right side upon page load. 

Image

Styling for the hero should be straightforward, probably can update global button styles to match the hero style , as there are not many other buttons on the page, and the button in the sticky CTA and header nav are already styled independently. This is the current hero block in EDS:

Image

roperev:
Logic to create button groupings and place image on the right. Animated hero image entrance from right.

Fix #97 

Test URLs:
- After: https://97-hero--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Desktop view:
image

Mobile view (550px):
image

### DIFF
--- a/blocks/hero/hero.js
+++ b/blocks/hero/hero.js
@@ -1,0 +1,75 @@
+export default function decorate(block) {
+  // Move the picture element to be positioned absolutely on the right
+  const picture = block.querySelector('picture');
+  if (picture) {
+    // Remove the picture from its parent div
+    const pictureParent = picture.parentElement;
+    pictureParent.remove();
+
+    // Append it directly to the hero block for absolute positioning
+    block.appendChild(picture);
+  }
+
+  // Group button labels with their buttons
+  // After removing picture parent, find the remaining content div
+  const contentDivs = Array.from(block.querySelectorAll('div > div'));
+  const contentDiv = contentDivs.find((div) => div.children.length > 0);
+
+  if (contentDiv) {
+    const allParagraphs = Array.from(contentDiv.querySelectorAll('p'));
+    const heading = contentDiv.querySelector('h1, h2');
+
+    // Find the label and button pairs
+    const label1 = allParagraphs.find((p) => !p.classList.contains('button-container')
+      && p.textContent.includes("Don't have"));
+    const button1Container = allParagraphs.find((p) => p.classList.contains('button-container')
+      && p.querySelector('.button.secondary'));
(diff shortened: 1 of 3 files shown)


blocks/hero/hero.js
```diff
@@ -0,0 +1,75 @@
...
+  const contentDivs = Array.from(block.querySelectorAll('div > div'));
+  const contentDiv = contentDivs.find((div) => div.children.length > 0);
+
+  if (contentDiv) {
+    const allParagraphs = Array.from(contentDiv.querySelectorAll('p'));
+    const heading = contentDiv.querySelector('h1, h2');
+
+    // Find the label and button pairs
+    const label1 = allParagraphs.find((p) => !p.classList.contains('button-container')
+      && p.textContent.includes("Don't have"));
```
helms-charity: This is too fragile. 
Do we need to create a formula for these? Maybe if we make the text that is to be associated with the button as a superscript, and in scripts.js we can create a button group based on `p>sup + p.button-container` siblings?

blocks/hero/hero.js
```diff
@@ -0,0 +1,75 @@
...
+    const label2 = allParagraphs.find((p) => !p.classList.contains('button-container')
+      && p.textContent.includes('Have an'));
+    const button2Container = allParagraphs.find((p) => p.classList.contains('button-container')
+      && p.querySelector('.button.primary'));
+
+    // Find footer paragraphs (the last two)
+    const footerParagraphs = allParagraphs.filter(
+      (p) => !p.classList.contains('button-container')
+        && !p.textContent.includes("Don't have")
+        && !p.textContent.includes('Have an'),
```
helms-charity: textContent is way too fragile

helms-charity (changes requested):

blocks/hero/hero.js
roperev: Yes, agree. I missed the hardcoded text that Cursor "helpfully" placed in the JS. I've refactored this to look for either `em > a` or `strong > a` patterns within the hero text and assemble the buttons based on that instead.

blocks/hero/hero.js
roperev: Definitely, my mistake. I've fixed this now.

helms-charity (approved):

Issue #76, opened 2025-11-14
Testimonials LD+JSON schema


helms-charity:
<img width="683" height="814" alt="Image" src="https://github.com/user-attachments/assets/97d0d23a-44b6-464f-b312-e3d933ab079e" />

this is added to the page head based on what's in the block.

PR #112, opened 2025-12-02
Add breadcrumb functionality for desktop sizes

### LINKED ISSUE #103: [block] Breadcrumbs
Most page types (specifically all credit card product pages, but also information pages like "Our History") look to have a simple breadcrumb navigation as the first section on the page. Per the AEM Live docs, this breadcrumb feature should be a part of `header.js`, however the boilerplate code in our repository contains nothing about breadcrumbs. See https://www.aem.live/developer/block-collection/breadcrumbs. 

Import the needed boilerplate code to our repo and add the breadcrumb functionality, as well as matching the styling shown below 

Note: there is an opacity value on the far right of the pill shaped background that lets some color come through on some pages like the `First Private` screenshot (last one) below, but not others, as it has a color of #E8E8E8 on the whole thing. It looks nice, we might as well do ours with the small opacity gradient too. 

- Image

- Image

- Image

- Image

- Image

roperev:
Hidden on mobile view sizes.

Fix #103 

Test URLs:
- After: https://103-breadcrumbs--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Additional Test URLs:
- After: https://103-breadcrumbs--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

image

image


blocks/header/breadcrumb-details.md
```diff
@@ -0,0 +1,257 @@
+# Breadcrumbs Implementation - Updated
```
helms-charity: what is all of this -- do we really need it? You can't speak to design decisions, only the customer can. And do we really need all the testing info for breadcrumbs? I would just add the bit about the "set breadcumbs=true if yuo want them" in the Library document and get rid of all this.

scripts/scripts.js
```diff
@@ -829,6 +829,177 @@ async function loadCategoryNav(main) {
...
+        pageTitle = pageTitle.split('|')[0].split(' - ')[0].trim();
+        crumbs.push({ title: pageTitle, url: currentUrl });
+      } else {
+        // For intermediate segments, convert path to readable title
+        let title = segment
+          .split('-')
+          .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+          .join(' ');
+        // Pluralize "card" to "cards" for better grammar in breadcrumbs
+        title = title.replace(/\bCard\b/g, 'Cards');
```
helms-charity, line 892: title.replace is OK for now. I have a feeling they will ask to be able to author a "breadcrumb title" on a page-by-page basis later.

helms-charity (commented):
Do you want to add LD+JSON schema with this PR? You can make a new ticket for it if you 'd rather.
image

It should be inline with the breadcrumbs instead of the head though. See my old project for inspiration. https://github.com/aemsites/888de/blob/954ec3ae06ab85327f64f2800efb5791d3e04384/scripts/scripts.js#L318-L336

blocks/header/breadcrumb-details.md
roperev: Great point, this is unnecessary. I'll remove it.

scripts/scripts.js
roperev: Would we prefer to add a "Breadcrumb title" field to UE right now and use normal og:title as the fallback instead?

helms-charity: Might as well, thank you.

roperev:
> Do you want to add LD+JSON schema with this PR? You can make a new ticket for it if you 'd rather. image
> 
> It should be inline with the breadcrumbs instead of the head though. See my old project for inspiration. https://github.com/aemsites/888de/blob/954ec3ae06ab85327f64f2800efb5791d3e04384/scripts/scripts.js#L318-L336

@helms-charity Should we add this schema addition to ticket #76? I imagine the process would be similar for both. Or if you want it done as part of this issue ticket, I'll add it for this PR instead.

roperev:
Final commit fixed all linting issues with `header.css` (~55 issues) except for descending specificity which has now been suppressed (8 warnings) for this file as the human readable order is preferred in this case.

component-models.json
```diff
@@ -2,6 +2,18 @@
...
     "id": "page-metadata",
     "fields": [
+      {
+        "component": "boolean",
+        "name": "breadcrumbs",
+        "label": "Show Breadcrumbs",
+        "description": "Show the breadcrumbs trail on the page (default: off)"
+      },
+      {
+        "component": "text",
```
helms-charity, line 12: You can add a "condition" to only show the breadcrumbsTitle if the show breadcrumbs is true. See https://github.com/aemsites/extweb-academy/blob/main/blocks/video/_video.json#L56-L67 as an example.

roperev: Great point, I should have thought of that. I've added it now.

helms-charity (approved):

PR #114, opened 2025-12-02
Add button group wrapper for responsive layouts

### LINKED ISSUE #106: Button groups
We need a way to define a button-group which consists of text and a button that should be wrapped together in a div. This way they can be 'moved' together when in a responsive layout.

Test page at https://main--idfc--aemsites.aem.page/library/buttons

I authored the text as Superscript.
So if part of our decorateButtons function we look for when the link has a parent P AND the previous sibling is p>sup, then wrap them both in a div class="button-group".

See https://github.com/da-pilot/sling/blob/main/eds/scripts/scripts.js#L179-L212 for inspiration, where I had a similar scenario, but in that case, I was looking for a button followed by subscript that was also linked.

zhummerz:
Wrap button containers with preceding superscript text in a 'button-group' div. Add corresponding CSS for alignment and spacing.

Fix #106 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/buttons
- After: https://106-btn-group--idfc--aemsites.aem.page/library/buttons

### DIFF
--- a/scripts/aem.js
+++ b/scripts/aem.js
@@ -441,6 +441,32 @@ function decorateButtons(element) {
       }
     }
   });
+
+  // Handle button groups: wrap button with preceding superscript text
+  // This allows text (in superscript) and button to be moved together in responsive layouts
+  element.querySelectorAll('p.button-container').forEach((buttonContainer) => {
+    // Check if this button container hasn't already been grouped
+    if (buttonContainer.parentElement?.classList.contains('button-group')) {
+      return;
+    }
+
+    // Check if the previous sibling is a <p> containing a <sup>
+    const previousSibling = buttonContainer.previousElementSibling;
+    if (previousSibling
+      && previousSibling.tagName === 'P'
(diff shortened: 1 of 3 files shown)


scripts/aem.js
roperev, line 1: Looks good, however we should put this in scripts.js rather than aem.js as we try to leave aem.js as untouched as possible for to streamline future baseline/boilerplate updates if they are needed. Can you move it there instead?

roperev (changes requested):

roperev:
One note, the before and after URLs in your PR description should both point to the same page, in this case your before URL should point to https://main--idfc--aemsites.aem.page/library/buttons, rather than the root page, as this will allow the PSI test step to run correctly and give accurate results.

scripts/aem.js
zhummerz: I moved the code from aem.js to stripts.js and corrected any linting errors.

roperev (approved):
Looks good now.

PR #139, opened 2025-12-08
Add modal block implementation with CSS and JavaScript

### LINKED ISSUE #128: autoblock modal pages
When a link is to a page under /modals/, open it in a modal.

See https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L593-L601 for how we did fragments.

Sling's version of the autoblocking modals:
https://github.com/da-pilot/sling/blob/main/eds/scripts/scripts.js#L167-L177

roperev:
https://www.aem.live/developer/block-collection/modal.

Add official modal block featuring styles for modal dialogs, includes a close button and functionality to close when clicking outside the dialog.

Be aware that the logic for triggering and auto-linking modals depends on the URL having the literal path `/modals/` somewhere within it. Ensure that naming of subfolders or a root global modals folders is the full `modals` literal text, and not `modal` or `global-modal` or `rupay-modals`, etc., as those will not get auto-modal'd. 

```
function autolinkModals(element) {
  element.addEventListener('click', async (e) => {
    const origin = e.target.closest('a');

    if (origin && origin.href && origin.href.includes('/modals/')) {
      e.preventDefault();
      const { openModal } = await import(`${window.hlx.codeBasePath}/blocks/modal/modal.js`);
      openModal(origin.href);
    }
  });
}
```

Fix #128

Test URLs:
- After: https://128-automodals--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/modal/modal.css
+++ b/blocks/modal/modal.css
@@ -1,0 +1,84 @@
+body.modal-open {
+  overflow: hidden;
+}
+
+.modal dialog::backdrop {
+  background-color: rgb(19 19 19 / 75%);
+}
+
+.modal dialog {
+  overscroll-behavior: none;
+  overflow-y: hidden;
+  width: calc(100vw - 48px);
+  max-width: 900px;
+  max-height: calc(100dvh - (2 * var(--header-height)));
+  padding: 0;
+  border: 1px solid var(--dark-color);
+}
+
+.modal dialog .modal-content {
+  box-sizing: border-box;
+  overflow-y: auto;
+  overscroll-behavior: none;
+  width: 100%;
+  max-height: calc(100dvh - (2 * var(--header-height)) - 48px);
+  padding: 24px;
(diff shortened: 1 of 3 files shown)


bunting-adbe (commented):
Don't we also need the function in the scripts.js that is mentioned in the documentation you included?
The block files look good to me.

roperev:
> Don't we also need the function in the scripts.js that is mentioned in the documentation you included? The block files look good to me.

Yes, we actually already had it in our version of scripts.js here: https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L66-L76. I just added a comment line above it to make it more consistent with the rest of the file. We just didn't have the actual js/css block code for the modal in our repo.

bunting-adbe (approved):
Sounds good. I didn't see any changes on that file, but that makes sense. 
Approved

PR #147, opened 2025-12-10
UE update for columns background image

### LINKED ISSUE #146: Columns "background" style
Similar to what was done for Sections, add a UE field to the Columns block so a "background-image" or "background" is authorable. Note that "background" replaces the "background-color"" in Sections so that gradients can be used.

Here are some examples on the Mayura page:

Image

They are both in the same section which has it's own background-image.

bunting-adbe:
Add UE configs to allow Columns block background image

Fix #146 

Test URLs:
- After: https://146-columns-background--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/columns/_columns.json
+++ b/blocks/columns/_columns.json
@@ -52,6 +52,14 @@
           "name": "rows",
           "value": "",
           "label": "Rows"
+        },
+        {
+          "component": "reference",
+          "valueType": "string",
+          "name": "columnsBackgroundImage",
+          "label": "Background Image",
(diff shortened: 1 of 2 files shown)


helms-charity (changes requested):
In your _columns.json -- you should have columnsBackgroundImage, columnsBackgroundImageMobile, (you don't need the Alt field for background images), and also add one more field for "columnsBackground" which you can label as "background color", but your JS will want to use the value for background=' ' (not background-image=' ')

helms-charity (approved):

PR #158, opened 2025-12-13
UE configs for Container

helms-charity:
Just the UE parts so I can author a page.
I set the component-filter on this, but not seeing it using "?ref" in the UE, so I think I need to merge it first.

for issue 153

Test URLs:
- After: https://153-container--idfc--aemsites.aem.page/


bunting-adbe (approved):
Looks good to me

PR #193, opened 2025-12-18
Create Table block UE

### LINKED ISSUE #191: Table with image options

bunting-adbe:
Create Table block and UE components.

Fix #191 

Test URLs:
- After: https://<branch>--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/table/_table.json
+++ b/blocks/table/_table.json
@@ -1,0 +1,281 @@
+{
+    "definitions": [
+      {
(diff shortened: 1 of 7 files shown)


bunting-adbe:
I took the table code from the MRV repo. Putting it here as a starting point, but will clean it all up for the requested needs.

roperev (approved):

PR #199, opened 2025-12-19
table styling

### LINKED ISSUE #191: Table with image options

bunting-adbe:
Table styling for Fees and charges section

Fix #191 

Test URLs:
- After: https://191-table-ue--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

image

### DIFF
--- a/blocks/table/_table.json
+++ b/blocks/table/_table.json
@@ -129,6 +129,14 @@
         },
         {
           "component": "reference",
+          "valueType": "string",
+          "name": "sectionBackgroundImage",
+          "label": "Background Image",
+          "description": "Background image of the section, displayed behind image options below",
+          "multi": false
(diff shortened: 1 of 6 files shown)


bunting-adbe:
found some issues. converting to a Draft and will finish up next week.

blocks/table/_table.json
```diff
@@ -127,6 +127,14 @@
             }
           ]
         },
+        {
+          "component": "reference",
+          "valueType": "string",
+          "name": "sectionBackgroundImage",
```
helms-charity: can you call it "tableBackgroundImage" or even just backgroundImage? (I think Evan named the one for Section).

helms-charity (approved):

PR #202, opened 2025-12-23
table styling finishes

### LINKED ISSUE #191: Table with image options

bunting-adbe:
Finished table styling. Now background color and image can be authored via the UE. Also added an update to the responsive background picture function in scripts, because it wasn't looking for mobile images. 
The modal also works (is authored) but still needs some styling work, both for the button and the rendering of the background picture. 

Fix #191 

Test URLs:
- After: https://191-table-ue--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

for modal see direct path:
https://191-table-ue--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura/modals/fee-and-charges-modal

image

image

### DIFF
--- a/blocks/table/table.js
+++ b/blocks/table/table.js
@@ -13,16 +13,8 @@ import { moveInstrumentation, createSource } from '../../scripts/scripts.js';
  */
 function extractImageUrl(element) {
   if (!element) return null;
-
-  if (element.tagName === 'IMG') {
-    return element.src;
-  }
-
-  if (element.tagName === 'PICTURE') {
-    const img = element.querySelector('img');
-    return img?.src || null;
-  }
-
+  if (element.tagName === 'IMG') return element.src;
+  if (element.tagName === 'PICTURE') return element.querySelector('img')?.src || null;
   return null;
 }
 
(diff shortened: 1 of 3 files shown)


blocks/table/table.js
```diff
@@ -69,65 +53,99 @@ function createResponsivePicture(desktopUrl, mobileUrl = null, alt = '') {
...
-      if (text && text !== 'true' && text !== 'false' && !text.match(/^\d+px$/)) {
-        imageAlt = text;
-        metadataCount += 1;
+  for (let j = metadataCount; j < rows.length; j += 1) {
+    if (rows[j]?.children.length === 1) {
+      const cell = rows[j].children[0];
+      if (!cell.querySelector('img, picture')) {
+        const text = cell.textContent?.trim();
+        if (text) {
+          const isColorPattern = text.startsWith('var(') || text.startsWith('#') || text.includes('gradient');
```
helms-charity, line 118: You've got some duplicate functionality, and I think it is a bit too complicated. For the background color option, we should set whatever is authored in there as "background" style (not background-image or background-color) like we did with the Section. I would like to reuse this part of the section code https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L766-L785 for that. We can go over it together when you are online.

helms-charity (changes requested):

helms-charity (approved):
Actually I decided to try something different for the Section style code after seeing the DA Author Kit, so we can merge your tables and revisit how to optimize the related functions later.

PR #226, opened 2026-01-08
fix for JSON

helms-charity:
1. I am putting back the section background property that I must have removed
2. Removing default values for the accordion link and linkText because they are not automatically wrtitten to the JCR upon creation, apparently. Now you have to author them, at which time they will be written back.

Test URLs:
- After: https://220-sectiongrids--idfc--aemsites.aem.page/


bunting-adbe (approved):
looks good

PR #232, opened 2026-01-08
Add delayed Google Tag Manager script loading to improve performance.…

### LINKED ISSUE #189: add GTM script
they use the same ID in staging/prod.
src="https://www.googletagmanager.com/gtm.js?id=GTM-M5CHMQ2Z"

Let's first try putting it in delayed.js like this:
https://github.com/helms-charity/mrvcopy/blob/main/scripts/delayed.js

zhummerz:
… Updated the ID to GTM-M5CHMQ2Z which matches in their current script.

Added the GTM script from the shared delayed.js script. The GTM ID was updated to match the same one used on Stage and Prod (M5CHMQ2Z). 

Fix #189 

Test URLs:
- After: https://189-add-gtm-script--idfc--aemsites.aem.page/

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -1389,6 +1389,33 @@ async function loadLazy(doc) {
  * Loads everything that happens a lot later,
  * without impacting the user experience.
  */
+
+// delayed GTM script
+
+window.dataLayer = window.dataLayer || [];
+
+function loadGTM() {
+  const gtmScript = document.createElement('script');
+  gtmScript.async = true;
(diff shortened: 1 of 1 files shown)


helms-charity:
@zhummerz I see that your code looks ok, but when I go to pages using your branch URL, the header is missing. Not sure if that has anything to do with your code or not, we will have to take a look.

helms-charity (approved):
I can see GTM loading so I'm approving. 

Image

I don't think the lack of header is related to this PR.

PR #240, opened 2026-01-09
accordion 1st item open, scroll

### LINKED ISSUE #236: accordion - 1st item starts opened
Should be like this when first on the page:

Image

Image

helms-charity:
- some style updates
- added feature so first accordion item is open by default
- added feature that moves the accordion open item to the top of the page, but didn't make it as annoying as the original :-D
- added code to footer.js so those don't get the same treatment.
- 
Fix #236-accordionopen

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/accordion
- After: https://236-accordionopen--idfc--aemsites.aem.page/library/accordion

### DIFF
--- a/blocks/accordion/accordion.css
+++ b/blocks/accordion/accordion.css
@@ -90,10 +90,14 @@
 }
 
 .accordion details summary {
+  position: relative;
+  padding-right: 46px;
+  cursor: pointer;
+  list-style: none;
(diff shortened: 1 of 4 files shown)


roperev (approved):
Looks like you've got all the styles dialed in now, seems great to me.

PR #243, opened 2026-01-09
new nav functionality

### LINKED ISSUE #225: Header refactor
https://stage.idfcfirst.bank.in/home-page-new

each section = top level?
inside,

bunting-adbe:
New navigation functionality - styling to come in another request
Review both desktop and mobile functionality.
I removed the header.css import of mob-header.css as that caused conflicts on the changes introduced. 

Fix #225 

Test URLs:
- Before: https://main--idfc--aemsites.aem.live/nav
- After: https://225-header-refactor--idfc--aemsites.aem.page/nav

### DIFF
--- a/blocks/header/header.js
+++ b/blocks/header/header.js
@@ -225,87 +225,89 @@ export default async function decorate(block) {
   navSectionsWrapper.appendChild(navSectionsUl);
   navSections.appendChild(navSectionsWrapper);
 
-  // Build nav-tools section from contentWrapper
+  // Build nav-tools section - look for tools/customer-service section
   const navToolsWrapper = document.createElement('div');
   navToolsWrapper.classList.add('default-content-wrapper');
 
-  // Find the search bar and other tools
-  const searchP = Array.from(contentWrapper.querySelectorAll('p')).find((p) => {
-    const strong = p.querySelector('strong');
-    return strong && strong.textContent.toLowerCase().includes('what are you looking for');
+  // Find the tools section (could be "tools", "customer service", "customer-service", etc.)
+  const toolsSection = Array.from(sections).find((section) => {
+    const sectionId = section.getAttribute('data-id');
+    return sectionId && (
+      sectionId.toLowerCase().includes('tool')
+      || sectionId.toLowerCase().includes('customer')
+      || sectionId.toLowerCase().includes('service')
+    );
   });
 
-  const specialP = Array.from(contentWrapper.querySelectorAll('p')).find((p) => p.textContent.trim() === "What's special about us" && !p.querySelector('strong'));
+  let searchP;
+  let toolsUl;
 
-  const toolsUl = Array.from(contentWrapper.querySelectorAll('ul')).find((ul) => {
-    const firstLi = ul.querySelector('li');
-    return firstLi && (firstLi.textContent.includes('Customer service') || firstLi.textContent.includes('Login'));
-  });
+  if (toolsSection) {
+    // Get content from the tools section
+    const toolsContent = toolsSection.querySelector('.default-content');
+    if (toolsContent) {
+      // Find search bar text
+      searchP = Array.from(toolsContent.querySelectorAll('p')).find((p) => {
+        const strong = p.querySelector('strong');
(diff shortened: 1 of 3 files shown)


blocks/header/header.js
```diff
@@ -157,148 +157,157 @@ export default async function decorate(block) {
...
-    } else if (child.tagName === 'P' && child.textContent.trim() === 'Investors' && !child.classList.contains('button-container')) {
-      // Investors section (p followed by ul)
-      currentLi = document.createElement('li');
-      currentLi.innerHTML = `<p><a href="https://www.idfcfirstbank.com/investors">${child.textContent}</a></p>`;
-      const nextChild = children[i + 1];
-      if (nextChild && nextChild.tagName === 'UL') {
-        currentLi.appendChild(nextChild.cloneNode(true));
-        i += 1; // Skip the next element
-      }
-      navSectionsUl.appendChild(currentLi);
```
roperev, line 205: Good for now since this is a step 1 of 2, but let's make sure we remember to cleanup debug output when we finish the step 2 tickets.

blocks/header/header.js
```diff
@@ -157,148 +157,157 @@ export default async function decorate(block) {
...
+  let toolsUl;
+
+  if (toolsSection) {
+    // Get content from the tools section
+    const toolsContent = toolsSection.querySelector('.default-content');
+    if (toolsContent) {
+      // Find search bar text
+      searchP = Array.from(toolsContent.querySelectorAll('p')).find((p) => {
+        const strong = p.querySelector('strong');
+        return strong && strong.textContent.toLowerCase().includes('what are you looking for');
```
roperev, line 252: This is a really fragile way to identify an element. Any chance we can locate it another way?

blocks/header/header.js
```diff
@@ -157,148 +157,157 @@ export default async function decorate(block) {
...
 
-    typeNextChar();
-  }
+  // Find the tools section (could be "tools", "customer service", "customer-service", etc.)
+  const toolsSection = Array.from(sections).find((section) => {
+    const sectionId = section.getAttribute('data-id');
+    return sectionId && (
+      sectionId.toLowerCase().includes('tool')
+      || sectionId.toLowerCase().includes('customer')
+      || sectionId.toLowerCase().includes('service')
```
roperev, line 238: This seems a bit prone to breaking if customer changes the section names/IDs. Any better way to identify these?

blocks/header/header.js
```diff
@@ -347,27 +356,225 @@ export default async function decorate(block) {
...
+        if (ul.children.length > 0) {
+          navSection.appendChild(ul);
+          navSection.setAttribute('data-fragment-loaded', 'true');
+          navSection.classList.add('nav-drop');
+
+          // Decorate any category-nav blocks that were added
+          const categoryNavBlocks = ul.querySelectorAll('.category-nav.block');
+          if (categoryNavBlocks.length > 0) {
+            // Dynamically import and decorate category-nav blocks
+            import('../category-nav/category-nav.js').then((module) => {
```
roperev, line 476: Relative path imports might break if we package/move content, maybe this should be a full path of `/block/category-nav/category-nav.js` instead? In theory it shouldn't since they're both blocks, but it feels like it would be better to give the full path reference here.

roperev (approved):
Left a few comments that you might want to address, but they're not dealbreakers, so I'm marking this as approved so Charity/Iustin aren't blocked over the weekend.

blocks/header/header.js
bunting-adbe: yep I usually do that, but as this is part 1 of several I went quick.

blocks/header/header.js
bunting-adbe: this is an update/modification to older code. I will see what I can do to improve.

blocks/header/header.js
bunting-adbe: this is an update/modification to older code. I will see what I can do to improve.

blocks/header/header.js
bunting-adbe: makes sense. I will do that.

PR #248, opened 2026-01-11
added the animation as an authorable video

iustinp:
Fix # issue 163

Test URLs:
- After: https://163-phone-animation--idfc--aemsites.aem.page/credit-card/rupay-credit-card


helms-charity (changes requested):
unfortunately, this video doesn't have a mediabus address, so I guess we need to look into what should happen in order for a mp4 to be treated like Franklin content.

helms-charity (approved):
Looks ok so far.

PR #250, opened 2026-01-12
styled metal card modal

### LINKED ISSUE #235: Authoring + style: Mayura popup
https://www.idfcfirst.bank.in/credit-card/metal-credit-card/mayura

Create/publish a new page under the existing mayura/modals/ directory

Image

 to create the content seen in this modal content:

Image

To test the popup itself, create (or update) a /library/modals page and just link to your new page. Clicking on that link should launch your modal page in a modal.

Then we will address the popup feature/timing from a specific page in another ticket https://github.com/aemsites/idfc/issues/234

iustinp:
Used rupay-credit-card page to test.
Not sure about why it's scrolling a bit, and the overall modal surrounding looks off, but i think that's part of another issue. 
Or have I missed something? I assumed I should not have to do anything to the modals in general.

Fix #235 

Test URLs:
- After: https://235-mayura-metal--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/modal/modal.css
+++ b/blocks/modal/modal.css
@@ -341,3 +341,179 @@ body.modal-open {
   margin: 0 0 25px;
   padding: 40px 0 0;
 }
+
+/* ===== Mayura Metal Card Modal Styles ===== */
+
+.modal .section[data-id="mayura-metal-card-modal"] {
+  position: relative;
+  font: 14px Inter, sans-serif;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:first-child > div {
+  align-items:flex-start;
+  height: auto;
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  gap: 20px;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:first-child > div > div:nth-child(2) {
+  padding-top: 10px;
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  justify-content: space-between;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:first-child > div h2 {
+  text-align: left;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:first-child > div a {
+    font-size: 14px;
+    line-height: 20px;
+    font-weight: 700;
+    color: #fff;
+    display: block;
+    padding: 13px 40px;
+    background: #9b1e26;
+    border-radius: 23px;
+    max-width: fit-content;
+    margin: 0;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:nth-child(2) > div {
+  margin-top: 40px;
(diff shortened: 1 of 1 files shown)


blocks/modal/modal.css
```diff
@@ -341,3 +341,179 @@ body.modal-open {
   margin: 0 0 25px;
   padding: 40px 0 0;
 }
+
+/* ===== Mayura Metal Card Modal Styles ===== */
+
+.modal .section[data-id="mayura-metal-card-modal"] {
+  position: relative;
+  font: 14px Inter, sans-serif;
```
helms-charity, line 349: use  --body-font-size-xs for font-size. Shouldn't need the Inter or sans-serif.

blocks/modal/modal.css
```diff
@@ -341,3 +341,179 @@ body.modal-open {
...
+    padding: 13px 40px;
+    background: #9b1e26;
+    border-radius: 23px;
+    max-width: fit-content;
+    margin: 0;
+}
+
+.modal .section[data-id="mayura-metal-card-modal"] > div > div:nth-child(2) > div {
+  margin-top: 40px;
+  display: -ms-grid;
```
helms-charity, line 387: remove all the -ms stuff.

blocks/modal/modal.css
```diff
@@ -341,3 +341,179 @@ body.modal-open {
...
+  display: none;
+  padding: 13px 40px;
+  background: #9b1e26;
+  border-radius: 23px;
+  max-width: -moz-fit-content;
+  max-width: fit-content;
+  margin: 20px auto 0;
+}
+
+@media (width <= 768px) {
```
helms-charity, line 471: All of this should be part of the initial styles. Use only < the breakpoints after that.

blocks/modal/modal.css
```diff
@@ -341,3 +341,179 @@ body.modal-open {
...
+.modal .section[data-id="mayura-metal-card-modal"] > .default-content a {
+  font-size: 14px;
+  line-height: 20px;
+  font-weight: 700;
+  color: #fff;
+  display: none;
+  padding: 13px 40px;
+  background: #9b1e26;
+  border-radius: 23px;
+  max-width: -moz-fit-content;
```
helms-charity, line 466: we don't need -moz definition either.

helms-charity (changes requested):
Check that the height of the modal is correct, I shouldn't have to scroll.
also can you author this on the Mayura page instead of the Rupay page?

iustinp:
Done everything except height&width + mobile-first, because:

Editing this is a bit difficult because of problems at the general modal level, so that's why i chose to do the desktop first approach (mobile is even worse to style..):
image

I therefore think that these should be fixed in a separate issue for all modal pop-ups, and then we can revise.

helms-charity (approved):
OK on your proposed changes. we can revisit the order of the CSS in a different ticket if needed.

PR #281, opened 2026-01-13
styled concept for metal card

### LINKED ISSUE #255: Heritage CC banner - Concepts tab block styling
Create a new page under /mayura/modals for this.
To author it, there is a new block to use called "tabs-cc-concept" created with a previous ticket, that has no styling.

In mobile, these "pointers" are like traditional tabs where the headings are on top.

Image

in desktop, the headings are on the right side...?

Image

JS and CSS to be added via the parent page or block which would handle the popup of this new fragment page.

iustinp:
Fix #255

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/test
- After: https://255-metal-card-concept--idfc--aemsites.aem.page/drafts/test

### DIFF
--- a/blocks/modal/modal.css
+++ b/blocks/modal/modal.css
@@ -519,21 +519,18 @@ body.modal-open {
   justify-content: center;
 }
 
-.section[data-id="mayura-metal-concept-2"] > div > div > div > div:first-child {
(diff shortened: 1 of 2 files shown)


helms-charity (approved):
THis works for me. In the future, if we need another one, we can separate out the specific concept JS to another file, but it's ok for now.

Issue #349, opened 2026-01-17
Header nav alignment 900-1200px


helms-charity:
On their site, https://stage.idfcfirst.bank.in/home-page-new, from 900px - 1200px wide the nav links are in a smaller font and closer together, and the search container is less wide so that all content still fits on 1 line.

Their header also is wider than ours in wide desktop size.

The space to the left of the logo is also smaller than ours, which helps too.

Ours should not go to 2 lines,
and in header.css line 908 we should set max-width: unset;

<img width="378" height="242" alt="Image" src="https://github.com/user-attachments/assets/34ba07de-d2fa-4160-a2fd-225fffd96e94" />

And between 900px - 1200px, we don't need the 70px right/left padding either.

<img width="1036" height="378" alt="Image" src="https://github.com/user-attachments/assets/5348746d-2aec-45c4-a307-4699795b588c" />

bunting-adbe:
I may have already addressed these as part of #350

helms-charity:
closed as the rest was handled via #356

PR #353, opened 2026-01-18
Header Nav Alignment

### LINKED ISSUE #350: Nav dropdown column widths - Tablet size
Between 900 - 1200px:
- the main nav links don't open on hover anymore, you have to click on them
- the columns of the dropdown content should be narrower, ours is overlapping as you get closer to 900px
- there's too much left spacing before our columns start
- the fonts should be smaller

Image

Image

bunting-adbe:
1. Found that the hover functionality doesn't work when you have Dev Tools Device Bar active. Not sure why, but it worked in every other scenario. I can dig further on that if wanted.
2. Column spacing is improved as you get to smaller sizes. It still gets messy below 950. I can try to make it work at that size also. Their break point is around 990 so they don't have that issue. 
3. Addressed the spacing on the left side, at smaller widths
4. Assigned smaller fonts at the tablet size. 

I noticed I may have also covered the items in #349 

Fix #350

Test URLs:
- After: https://350-nav-tablet--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -250,7 +250,7 @@ header nav .nav-sections .nav-fragment-section-title {
   padding: 14px 40px 14px 20px;
   margin: 0;
   font-weight: 600;
-  font-size: 16px;
+  font-size: var(--body-font-size-s);
   line-height: 1.4;
   color: #353535;
   background-color: #f9f9f9;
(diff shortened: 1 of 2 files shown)


helms-charity (approved):
it looks much better in tablet size now. Thanks

PR #379, opened 2026-01-20
UE only: Adding JSON page properties

### LINKED ISSUE #132: Page metadata for SEO, Search
Universal Editor _page.json needs to be updated, but I don't think we need all of their current SEO fields:
Rupay card example:
Image
The "script" area seems to be a custom LD+JSON schema just pasted in. I don't know what (if anything) is being hidden when "hide schema" is checked. Compare with the Mayura page where the box isn't checked and they also have pasted in "script".

Image

They also need a cq:tags picker:

Image

We can add some of their Search meta items too, though I don't know that we will actually use any of these for our POC pages.

Image

THere is also this tab of 'additional info', but it's blank for Rupay and Mayura.

Image

helms-charity:
This ticket takes into account certain page properties that IDFC currently has says they need to author, and I've matche up with the JCR property name value where possible. Some of these don't do anything except get added to the page metadata, others will need some JS implementation, to come in other tickets.

NOTE: There is an error for "avoid using orphans" for h1Title, but I want this field to match the JCR so their reporting doesn't break, so we can ignore that.

Fix #132

Test URLs:
- After: https://132-pageproperties--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/component-models.json
+++ b/component-models.json
@@ -3,10 +3,33 @@
     "id": "page-metadata",
     "fields": [
       {
+        "component": "tab",
+        "label": "Page meta + configs",
+        "name": "tab1"
+      },
+      {
         "component": "boolean",
         "name": "breadcrumbs",
         "label": "Show Breadcrumbs",
         "description": "Show the breadcrumbs trail above page content"
+      },
+      {
+        "component": "boolean",
+        "name": "analytics-disable",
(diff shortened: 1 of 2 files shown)


techmaven (approved):
Verified fields in UE

PR #383, opened 2026-01-20
remove beginning slash on folder

### LINKED ISSUE #132: Page metadata for SEO, Search

helms-charity:
Hoping this change will fix this console issue
image

Fix #132 

Test URLs:
- After: https://132-pageproperties--idfc--aemsites.aem.page/

### DIFF
--- a/component-models.json
+++ b/component-models.json
@@ -80,7 +80,7 @@
         "name": "tags",
         "valueType": "string",
         "label": "Keywords/Tags",
-        "rootPath": "/content/cq:tags"
+        "rootPath": "content/cq:tags"
       },
       {
         "component": "text",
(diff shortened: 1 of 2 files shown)


sthirumal (approved):
Approve the change

PR #384, opened 2026-01-20
Not Draft anymore

iustinp:
UPDATE: This is ready for review, since I think everything is done. 
----------------------
For ease of editing, I've temporarily disabled autohide on mouseleave in header.js as well as the css hover hiding. 

What is done so far:
- Authored the first 2 categories in Credit Cards
- Created new image field for these new cards
- Created no-background option on category-nav items
- Mostly styled the cards column
- Authored in only one card from the second category (from the right of the nav dropdown)

What more is needed:
- Author in and place the card on the right of the dropdown - it should already be styled 
- Author in the rest of the right-side cards
- Javascript to bring in the right-side cards if they exist and place it appropriately
- Check that styling works as does for the other sections, other than Credit Cards. 
- Remove the temporary hide disable code. 

Known issues: Messed up some stuff on the mayura sub-nav. 

If no one picks it up, I will continue in the morning. 

Fix #issue 362

Test URLs:
- After: https://362-cc-nav-fragment--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/category-nav/category-nav.css
+++ b/blocks/category-nav/category-nav.css
@@ -167,7 +167,7 @@
 
 .category-nav-card .category-nav-card-link {
   display: flex;
-  flex-direction: column;
+  flex-direction: column-reverse;
   position: relative;
   padding: 12px 14px 14px;
   border-radius: 8px;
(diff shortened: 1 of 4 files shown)


helms-charity (approved):
Looking good, I fixed the duplicate 'block' declarations.

PR #386, opened 2026-01-21
Fix 2nd fee modal background

### LINKED ISSUE #371: Mayura "fees and charges" modal x2
Setting this as "high", as no doubt the agency will call it out in tomorrow's QA.

Currently there are 2 links which launch differently-styled modals.
Can we try to get the second link to open in the same style as the first?

First link:

Image
opens up the nicer looking modal

Image

Second link:

Image

opens the modal with the white background, which we don't want.

Image

techmaven:
Fix background of Fees and charges View More dialog.

Fix #371 

Test URLs:
- After: https://371-fees-modal--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/modal/modal.js
+++ b/blocks/modal/modal.js
@@ -454,9 +454,14 @@ export async function openModal(fragmentUrl, options = {}) {
   const path = fragmentUrl.startsWith('http')
     ? new URL(fragmentUrl, window.location).pathname
     : fragmentUrl;
-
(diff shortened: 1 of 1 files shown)


techmaven:
The table inside the dialog is still getting a slightly different background; another style attribute being added directly to the table-container div. I ran out of time to figure out where that is coming from.

helms-charity (approved):
This is fine, the extra background image can be fixed via authoring.

PR #388, opened 2026-01-21
Table: Add left and right padding

### LINKED ISSUE #367: Table padding left and right
Table:
The table block should have 16px padding so data isn't up against the edge of it, which is especially noticeable if a table backrgound color is authored.

See the /library/table page, also

Image
(modal is from the Mayura page) https://main--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

amarghioali:
Fix #367 

Added 16px horizontal padding (left and right) to the table within the "Fees and Charges" popup window.

The remaining styling for the popup window will be addressed in a separate issue or PR.

Test URLs:
- After: https://367-table-padding--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/table/table.css
+++ b/blocks/table/table.css
@@ -7,6 +7,10 @@
   position: relative;
 }
 
+.table-wrapper{
+  padding: 0 16px;
+}
+
 .table-background-image {
   position: absolute;
   top: 0;
--- a/blocks/table/table.css
+++ b/blocks/table/table.css
@@ -7,7 +7,7 @@
   position: relative;
 }
 
-.table-wrapper{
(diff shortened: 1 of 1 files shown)


helms-charity (commented):
This fix is not just for the modal, it is for all tables. See library page at the bottom - should also have the visible padding. https://367-table-padding--idfc--aemsites.aem.page/library/table

helms-charity (approved):
This fixes the immediate problem at hand, now that I see the "fees-and-charges" specific styling is what broke it.... we can deal with that later.

Image

PR #416, opened 2026-01-22
Fix Hero-HeritageCC display in UE

roperev:
Fix Hero-HeritageCC display in UE, add new styling view to be able to understand and author a hero-heritage cc component and not have the entire thing barf out all of its components at the top of the page, or render the hero with infinite height which breaks authoring entirely. 

NOTE: Currently after fixing the component "name" property value in CRXDE so the block code loads instead of 404s, the height of that component in UE becomes infinite. As such, this PR should be merged as soon as possible. 

image

image

Test URLs:
- After: https://hero-ue-display-fix--idfc--aemsites.aem.page/


techmaven (approved):
Looks good.

PR #417, opened 2026-01-22
Add reward points table styles and improve table rendering logic

### LINKED ISSUE #408: Authoring - Needed card to be clickable and open other page for information - Rewards Points
Original idfc

Image

This are missing in ours

Image

On clicking each card should show below

Image

Image

sthirumal:
- Introduced specific CSS styles for the reward points table, including layout and padding adjustments.
- Updated JavaScript to conditionally create table headers and filter out empty rows for better data presentation.

Fix #408

Test URLs:
- After: https://408-reward--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/table/table.css
+++ b/blocks/table/table.css
@@ -116,10 +116,60 @@
   background-color: var(--white);
 }
 
+/* Reward Points Table Specific Styles */
+.table.block#reward-points {
+  margin: 0;
+  padding: 0;
+}
+
+.table.block#reward-points table {
+  border: 1px solid var(--white);
+  border-radius: 8px;
+  overflow: hidden;
+  border-collapse: separate;
+  border-spacing: 0;
+}
+
+.table.block#reward-points table td {
+  padding: 10px 12px;
+  border: 1px solid var(--white);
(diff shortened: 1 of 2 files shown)


sthirumal:
My comments :

Initially I though this only required authoring. Then later realized that it needs table and also table styling.

I would like to mention few items 

- Looks like UE has bug for the table component. When I added table somehow its adding few empty rows. And in DOM it can be easily see that empty rows are introduced.
   I tried multiple times adding table component every time it introduces empty rows at beginning.
  
   As workaround I had to add logic in table.js to check for the empty row and remove/trim it. 

    Thats why need you experts insite on this code change. 


- In order to add styling of table i have used the "reward-points" variant. And the table.css has stying done for table using the "reward-points" variant.

  Below is the outcome of the table style. Probably it will need some more fine tuning.

image

image



- And the buttons at the bottom needs styling; wondering if this is priority to take care or can be considered as LOW priority.

helms-charity (approved):
You did a nice job, and yes there is a table bug with its own ticket already.

PR #419, opened 2026-01-22
feat(category-nav): add Cards block support with click-to-open dropdown

### LINKED ISSUE #391: Missed req: Bell + dropdown on category nav
See https://www.idfcfirst.bank.in/credit-card/rupay-credit-card
They have one more item in their category nav (which only shows in desktop).

Image

We need to add this to our Category Nav block. Although the live site only has the bell on Rupay, let's keep it on both. We can always CSS hide it in the Mayura template if they complain.

To start us off, I've created a copy of the Category Nav (called ccnav2, "Category nav with bell") in the /framework folder 

Image

and added 2 "blog" cards (which are the only cards style that is allowing 2 links in the text).

Image

### LINKED ISSUE #421: Refactor Navigation Menu Layout (ccnav) to Match IDFC
The current layout of the secondary navigation bar (ccnav) differs from the IDFC reference site. Specifically:

- IDFC Site: Features a split layout with the navigation menu items aligned to the left and the notification bell icon positioned on the far right.
- Current Project: All items are currently grouped together in the center.

### Technical Changes Implemented
To facilitate this layout transition, we are using a specific semantic CSS class for the bell icon: .category-nav-bulletin-card-notification.

By targeting this class, we can decouple the bell icon's positioning from the main navigation list. This allows the icon to be pushed to the far right independently while maintaining the menu items on the left.

Ref: https://github.com/aemsites/idfc/pull/419

### Task Requirements
The goal of this issue is to adjust the CSS and layout to achieve the following:
- Alignment: Push the regular menu items to the left and the bell icon to the far right.
- Width: Match the overall container width of the IDFC reference site (spanning from the IDFC logo edge to the Login button edge).
- Consistency: Ensure the horizontal spacing and padding match the reference design.

Image

amarghioali:
Fix #391
Fix #421  

Add functionality to display Cards block items in category-nav dropdown menu with enhanced UI and interaction patterns.

Features implemented:
- Parse sections containing Cards blocks and integrate into category navigation
- Support icon-based menu items (e.g., :bell: icon) using icon decoration
- Add click-to-open behavior for Cards block dropdowns (vs hover for regular items)
- Add close button (×) in top-right corner of dropdown
- Add navigation arrows with pagination (max 3 cards visible, disabled when ≤3)
- Extract and display dropdown container title from second text element

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/rupay-credit-card1
- After: https://391-bell--idfc--aemsites.aem.page/drafts/rupay-credit-card1
- Reference: https://www.idfcfirst.bank.in/credit-card/rupay-credit-card


roperev (changes requested):
Bogdan, 

I have a couple of main issues with this change:

1. I think maybe restyling the entire top nav positioning type has broken the normal card layouts and styling (doesn't flex wrap nicely anymore):

Image

I'd suggest that rather than using a category-nav-left and category-nav-right, we just add the bell/notification item and logic as the last normal li item (you're doing this at line 555 but to the navListRight/navListLeft containers already) to the main category nav, and just use a more simple child selector (like a `.top-nav-left>li:nth-last-child(1) { margin-left: auto; }` to shift it all the way over to the right (as far as you want). You could also just give it a semantic CSS name like `category-nav-bulletin-card-notification` or something. This would be a lot less styling code to shift the whole nav out of `block` to `flex`, then most of it back to `block` again (in terms of lines of code on a bunch of selectors)

If we need to adjust the positioning of the category-nav to be more to the left side rather than fully centered, that could be done independently of the position of the right bell icon if we did it as described above. 

Either way, we need to fix the broken cards. Let me know if you have any questions on this. 

2. Secondly, it looks like you built an entire js based card carousel feature for this cards display. It looks good and functions alright, but we're already using the full swiper.js library (at `/scripts/swiperjs/swiper-bundle.min.js`) for all card movement logic in the whole site. 

You can just ask cursor to refactor your custom carousel function to use that swiper library instead, you may have to direct it a little to keep your next/previous arrows, but that will unify the feel of your cards and ensure broad compatibility without bugs on all devices. This saves ~600 lines of js code since swiper is already loading for every page anyway. 

3. It appears your code is reading the cards' details from out of the content on the authored page blocks, at least it's authored in the page on your `/drafts/rupay-credit-card1` page, and I can't get it to appear on something like https://391-bell--idfc--aemsites.aem.page/credit-card/rupay-credit-card at all. This content needs to be authored either as it's own fragment page within `/framework/` or specifically at the end of `/framework/ccnav`, which is where the rest of that category-nav content is currently authored. I believe Charity did some of the legwork for this on `/framework/ccnav2`, you probably already took a look at it there.

amarghioali:
Hi @roperev ,
Thanks for the review! I’ve addressed the first two items—please take a look when you have a moment and let me know your thoughts.

Regarding item #3: I am using the framework implementation, specifically a copy of ccnav (ccnav2) created by Charity. I’m working within this copy to experiment safely without risking any breaking changes to the existing functionality.
image

Currently, the bell icon is only authored in /framework/ccnav2. Pages using ccnav or ccnav1 do not have the bell icon yet; it simply needs to be added to those versions following the same authoring logic used in ccnav2.
image

For example, the page /credit-card/rupay-credit-card is currently using ccnav1. If we switch it to ccnav2, the bell icon should appear as expected.
image

I haven't made any changes in AEM yet to avoid interference, especially since I noticed on Slack that there are currently some publishing issues. Please test this setup when you have a few minutes.

Thanks

cc: @helms-charity

blocks/category-nav/category-nav.js
```diff
@@ -379,13 +799,48 @@ export default function decorate(block) {
   // Find ALL category-nav blocks on the page
   const allCategoryNavBlocks = document.querySelectorAll('.category-nav.block');
```
helms-charity: We can assume there is only 1 and we're inside the category-nav.js already.

blocks/category-nav/category-nav.js
```diff
@@ -379,13 +799,48 @@ export default function decorate(block) {
...
 
-  if (allCategoryNavBlocks.length === 0) {
+  // Also find sections with Cards blocks (not category-nav blocks)
+  const allSections = document.querySelectorAll('.section');
```
helms-charity: you're already in category-nav.js, so you don't want to do a document.querySelectorAll for anything (including all sections). You can reduce the scope of this "allSections.forEach".

helms-charity (changes requested):

blocks/category-nav/category-nav.js
```diff
@@ -218,17 +312,82 @@ export function buildDropdown(categoryData) {
...
+
+    carouselWrapper.appendChild(menuCardList);
+
+    // Always add navigation arrows for Cards block items
+    const navigationWrapper = document.createElement('div');
+    navigationWrapper.classList.add('category-nav-navigation');
+
+    const prevButton = document.createElement('button');
+    prevButton.classList.add('category-nav-nav-button', 'category-nav-nav-prev');
+    prevButton.setAttribute('aria-label', 'Previous cards');
```
helms-charity: Just change to "Previous" for the aria-label

blocks/category-nav/category-nav.js
```diff
@@ -218,17 +312,82 @@ export function buildDropdown(categoryData) {
...
+    prevButton.innerHTML = '<span class="nav-arrow">‹</span>';
+    prevButton.disabled = true; // Start disabled
+
+    const nextButton = document.createElement('button');
+    nextButton.classList.add('category-nav-nav-button', 'category-nav-nav-next');
+    nextButton.setAttribute('aria-label', 'Next cards');
```
helms-charity: "Next" is all we need for the aria-label

blocks/category-nav/category-nav.js
```diff
@@ -254,19 +414,59 @@ function buildUnifiedNavigation(categoriesData) {
...
-      const target = document.getElementById(category.id);
-      if (target) {
-        const yOffset = -200;
-        const y = target.getBoundingClientRect().top + window.pageYOffset + yOffset;
-        window.scrollTo({ top: y, behavior: 'smooth' });
-      }
-    });
+
+    // Different click behavior for cards block items vs regular items
+    if (category.isFromCardsBlock) {
```
helms-charity, line 430: Can you combine this IF with the one at line 418?

blocks/category-nav/category-nav.js
```diff
@@ -327,6 +540,218 @@ function positionDropdowns() {
...
+
+  // Extract content from card body
+  const cardBody = cardLi.querySelector('.cards-card-body');
+  let title = '';
+  const paragraphElements = [];
+
+  if (cardBody) {
+    // Check for heading first
+    const heading = cardBody.querySelector('h1, h2, h3, h4, h5, h6');
+    const paragraphs = Array.from(cardBody.querySelectorAll('p'));
```
helms-charity: we already check for this in the cards.js https://github.com/aemsites/idfc/blob/main/blocks/cards/cards.js#L301

blocks/category-nav/category-nav.js
```diff
@@ -327,6 +540,218 @@ function positionDropdowns() {
...
+}
+
+/**
+ * Parse a section with a Cards block to extract category data
+ * @param {HTMLElement} section - The section element containing Cards block
+ * @returns {Object} Category data object
+ */
+function parseSectionWithCardsBlock(section) {
+  // Get text elements outside of blocks
+  const allTextElements = section.querySelectorAll('p, h1, h2, h3, h4, h5, h6');
```
helms-charity: You're doing this at 563 too

blocks/category-nav/category-nav.js
```diff
@@ -379,13 +804,40 @@ export default function decorate(block) {
...
+    const firstNavBlock = allCategoryNavBlocks[0];
+    const container = firstNavBlock.closest('.section')?.parentElement;
+
+    if (container) {
+      const allSiblings = Array.from(container.children).filter((el) => el.classList.contains('section'));
+
+      allSiblings.forEach((section) => {
+        const hasCardsBlock = section.querySelector('.cards.block');
+        const hasCategoryNav = section.querySelector('.category-nav.block');
+        const hasIcon = Array.from(section.querySelectorAll('p, h1, h2, h3, h4, h5, h6')).some(
```
helms-charity: this is the 3rd time you are looking for these elements, so if you really need to do this query, maybe make a helper function to use.

blocks/category-nav/category-nav.js
```diff
@@ -327,6 +540,218 @@ function positionDropdowns() {
...
+  if (textElementsOutsideBlocks.length > 0) {
+    const firstElement = textElementsOutsideBlocks[0];
+    const icon = firstElement.querySelector('span.icon');
+    if (icon) {
+      // Get the icon name from classes like "icon-bell"
+      const iconClass = Array.from(icon.classList).find((c) => c.startsWith('icon-'));
+      let iconName = iconClass ? iconClass.substring(5) : 'icon';
+
+      // For Cards blocks, always use bell-outline icon instead of regular bell
+      if (iconName === 'bell') {
```
helms-charity: This makes no sense. Just use the icon you added (bell-outline). If they want to change it, they can.

amarghioali (commented):
Hi @helms-charity 
I've updated the PR based on your comments. Please take another look when you have a chance. Thanks!

helms-charity (approved):
It looks better than the original. Sometime we should revisit the entire category-nav to slim down the JS, but for now it's good.

PR #422, opened 2026-01-22
Hide CTA popup at wide width

### LINKED ISSUE #405: sticky CTA - mobile only Mayura
Needs authoring, then some code to hide it....
can put some CSS in the /templates/mayura.css file for a quick win.

24. Sticky CTA is missing in mobile view

idfc orginal

Image

Our site 

Image

techmaven:
Hide CTA popup when width >= 990

Fix #405

Test URLs:
- After: https://405-sticky-cta--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/templates/mayura/mayura.css
+++ b/templates/mayura/mayura.css
@@ -83,6 +83,21 @@ color: var(--dark-red-color);
   line-height: 1.5;
 }
 
+.mayura .sticky-cta.block {
+  background: rgba(239, 239, 239, 0.6)
+}
+
+.mayura .sticky-cta-subtitle-content {
+  color: rgb(26, 26, 26);
+  padding-top: 10px;
+  font-size: 14px;
+}
+
+.mayura .sticky-cta-container .sticky-cta-content a.button {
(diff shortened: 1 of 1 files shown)


helms-charity (commented):
You did exactly what was asked, but I think the customer changed it on us, so I'm going to request changes.

- instead of hiding it at >=990px, now it is hidden wider than >=768px.
- Button only says "Apply now", has arrow-right-white icon added and is silver... (it is actually the same as the topmost button next to "The Concept" so you can reuse that style).
- Text under the button is now "Joining fee: ₹5,999+GST and Annual Fee: ₹5,999+GST (2nd year onwards)". I can see that text starts disappearing on their site at around 554px wide, but I think that's a bug, so ignore that.
- the background of sticky CTA is a translucent light instead of dark
- I also see that their CTA doesn't show up until you've scrolled down the page, but ignore that for now too. I'll need to get more direction from the customer.

techmaven:
Changes completed as discussed.

helms-charity (approved):
looks good

PR #430, opened 2026-01-23
UE only -- changed fields to be separate

helms-charity:
Using underscores in my last PR was a bad choice, because both mobile and desktop text was in the same cell.

image

Fix 428

Test URLs:
- After: https://428-stickynavredo--idfc--aemsites.aem.page/


techmaven (approved):
Looks good

PR #432, opened 2026-01-23
Add full search functionality

### LINKED ISSUE #418: Enable 'phase 1' site search for go live
I learned from the agency lead that they have a robust Solr search solution, with AI and all the bells and whistles. He can give me API details to (eventually) implement it on our EDS pages.

For the short term, I've learned there is a dedicated search page at https://www.idfcfirst.bank.in/search we can leverage.
He said they pass in some parameter from the page to Solr, so I want to redirect clicks on our search field to the search page with the parameters from each page (if we can find them by inspecting the network tab?).

roperev:
Add full search functionality with new results display and fallback suggestions until real API is hooked up.

- Update CSS for dropdown overlays and search results styling, ensuring a cohesive user experience.
- Refactor JavaScript to manage search input, results fetching, and default content restoration effectively.

Note that this is using placeholder link content (line 558-614) to populate the search results which I just generated from our top level nav and our ccnav category nav framework pages.

Upon pressing `enter` or clicking the `View all results for <term>`, the user is redirected to IDFCs current search page with a format of `https://www.idfcfirst.bank.in/search?skey=<term>`. This *should* work, however their search feature is broken and doesn't actually take in the query parameters to execute the search....so if they fix that in their code, our external "redirect" to full search will work. 

API integration will need some work, see lines 667-693, when/if we do have access to something from their SOLR search at some point. 

Fix #418

Some screenshots of empty search results, single result, multiple result output styling:
image

image

image
Above image shows hover-over effect on one of the results (Ashva card). 

Test URLs:
- After: https://418-search--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/header/header.js
+++ b/blocks/header/header.js
@@ -529,6 +529,253 @@ export default async function decorate(block) {
       closeDelay: 100,
     });
 
+    const searchInput = searchBox.querySelector('.search-input');
+    let searchTimeout = null;
+    let defaultDropdownContent = null;
+
+    // Store the default dropdown content (Most Searched, Top Products, Accounts)
+    const storeDefaultContent = () => {
+      if (!defaultDropdownContent && search.dropdown.querySelector('.default-content')) {
+        defaultDropdownContent = search.dropdown.querySelector('.default-content').cloneNode(true);
+      }
+    };
+
+    // Restore default dropdown content
+    const restoreDefaultContent = () => {
+      // Remove search results
+      const searchResults = search.dropdown.querySelector('.search-results');
+      if (searchResults) {
+        searchResults.remove();
+      }
+
+      // Show the first section again
+      const firstSection = search.dropdown.querySelector('.section');
+      if (firstSection && firstSection.style.display === 'none') {
+        firstSection.style.display = '';
+      }
+    };
+
+    // Get fallback suggestions based on query
+    const getFallbackSuggestions = (query) => {
+      const allSuggestions = [
+        // Popular Products & Services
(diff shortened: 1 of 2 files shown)


blocks/header/header.js
```diff
@@ -529,6 +529,253 @@ export default async function decorate(block) {
...
+      if (firstSection && firstSection.style.display === 'none') {
+        firstSection.style.display = '';
+      }
+    };
+
+    // Get fallback suggestions based on query
+    const getFallbackSuggestions = (query) => {
+      const allSuggestions = [
+        // Popular Products & Services
+        { title: 'Personal Loan', url: '/personal-banking/loans/personal-loan', type: 'Loan' },
```
helms-charity, line 562: These URLs need to go to the live site. Might be easiest to add a new variable for "domain" or something, then add $domain in front of all these paths?

blocks/header/header.js
```diff
@@ -529,6 +529,253 @@ export default async function decorate(block) {
...
+    const displaySearchResults = (results, query, container) => {
+      const loadingEl = container.querySelector('.search-results-loading');
+      if (loadingEl) {
+        loadingEl.remove();
+      }
+
+      if (results.length === 0) {
+        container.innerHTML += `
+          <div class="search-results-empty">
+            <p>No results found for "${query}"</p>
```
helms-charity, line 632: Instead of printing text when results.length === 0, can you do an immediate redirect to their search page? 
Because otherwise if you search for any popular thing that's not in your list (like "investors") it appears to the user that the bank has no investors....

helms-charity (approved):
I want to merge this in, so I've created https://github.com/aemsites/idfc/issues/435 for addressing my comments.

PR #434, opened 2026-01-24
New block to replace sticky-cta

### LINKED ISSUE #428: Sticky CTA refactor
We can simplify this. Authors can create either a primary or secondary button, and add an icon using colon notation if they want. The background color can be controlled by the section instead of the component.

Image

"CTA text" should be used for all widths unless the CTA Mobile Text is authored.

Not sure exactly how this field is used, and I don't see the text on the Rupay page HTML.

Image

helms-charity:
Once approved, we will need to re-author the 2 pages to use this block instead of "sticky-nav", and we can remove the old block code.

This block has 24 lines of JS, and 86 lines of CSS.
Old block has 64 lines of JS, and 146 lines of CSS!

Fix #428 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/cta-sticky
- After: https://428-stickynavredo--idfc--aemsites.aem.page/library/cta-sticky

Live pages for comparison:
https://www.idfcfirst.bank.in/credit-card/rupay-credit-card
https://www.idfcfirst.bank.in/credit-card/metal-credit-card/mayura (only in mobile)

### DIFF
--- a/blocks/cta-sticky/_cta-sticky.json
+++ b/blocks/cta-sticky/_cta-sticky.json
@@ -25,14 +25,16 @@
           "valueType": "string",
           "name": "ctaText",
           "label": "CTA Text",
-          "value": ""
+          "value": "",
+          "description": "Only visible on desktop"
         },
         {
           "component": "richtext",
           "valueType": "string",
           "name": "ctaTextMobile",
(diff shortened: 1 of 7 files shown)


bunting-adbe (approved):
looks great.

PR #444, opened 2026-01-26
big font cleanup and misc

### LINKED ISSUE #398: Global font styles - cleanup
We can use style tokens/variables for the all-encompassing "font" property (see https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font).

Units of measure:
- font sizing in REM
- font-related spacing in EM
- Browser root size is 16px by default, so don't declare it (lets people reset their root text size for accessibility).

This ticket is very hard, because the customer has been able to choose any font size they want via the RTE, and have individual page CSS too which affects the fonts.

We need to see if we can determine what their default/global font sizes, weights, spacing, and line-heights are for mobile and go from there.

These font styles also need to be all using style tokens, so that they can easily be overridden in the Design sheet.

helms-charity:
convert to REM units, replace most px fonts. Also changed out the font files to smaller ones.
Renamed a couple gradients.

Fix #398

Test URLs:
- After: https://398-fonts--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

- After: https://398-fonts--idfc--aemsites.aem.page/credit-card/rupay-credit-card


styles/styles.css
```diff
@@ -31,9 +39,9 @@
   --gray-btn-dark: #4b4b4b;
   --gradient-translucent: linear-gradient(180deg, rgba(255, 255, 255, 0.32) 0%, rgba(255, 255, 255, 0.24) 100%);
   --gradient-pearl: linear-gradient(73deg, #C7C6C6 -74.4%, #F1F1F1 23.15%);
-  --gradient-upi-steps-red: linear-gradient(165deg, #ca0919 0%, #8C2034 40%, #732C53 80%, #513873 100%);
-  --gradient-tabs-upi-link-red: radial-gradient(at top left, #AC2A34 0%, #661E2A 50%, #140F23 100%);
-  --gradient-important-documents: linear-gradient(to bottom, #C5767D, #9E1E29);
```
roperev, line 36: I think these are being referenced by authored gradient color names for these various section, as section background gradient styles. It looks like you've already updated them.

styles/styles.css
```diff
@@ -77,23 +86,24 @@
...
-  --heading-font-size-l: 34px;
-  --heading-font-size-m: 27px;
-  --heading-font-size-s: 24px;
-  --heading-font-size-xs: 22px;
+  --heading-font-size-xxl: 2.8125rem; /* 45px */  
+  --heading-font-size-xl: 2.25rem; /* 36px */
+  --heading-font-size-l: 2rem; /* 32px */
+  --heading-font-size-m: 1.75rem; /* 28px */
+  --heading-font-size-s: 1.5rem; /* 24px */
+  --heading-font-size-xs: 1.375rem; /* 22px */
```
roperev, line 107: Any reason the xxl/xl sizes are ~10px smaller than previously?

templates/mayura/mayura.css
```diff
@@ -90,30 +93,38 @@ color: var(--dark-red-color);
...
-  font-size: 14px;
+  font-size: var(--body-font-size-s);
 }
 
 .mayura .sticky-cta-container .sticky-cta-content a.button {
   background: linear-gradient(270deg, #332B28 0%, #514A44 21.11%, #8B837D 50.14%, #625A50 71.74%, #7A766D 90.18%, #433C32 100%);
   border: none;
 }
 
+@media (width >= 768px) {
```
roperev, line 104: Is this our tablet breakpoint for these or should this be 600px to fall off after mobile size breakpoint?

blocks/footer/footer.css
```diff
@@ -1,17 +1,9 @@
-
```
roperev, line 1: Lots of changes through this one, looks very slightly different when comparing before/after but still looking good and cutting over 100 lines is awesome.

blocks/hero-heritage-cc/hero-heritage-cc.css
```diff
@@ -100,7 +100,7 @@
 .hero-heritage-cc-header-cta a.button.primary {
```
roperev, line 100: I think this file has too many variations from the original look of the big Mayura Hero in terms of text sizes and positioning, see screenshots:

Image

Image

roperev (approved):
Everything seems totally fine here except for the big Mayura hero text sizing and spacing, which I noted. I think those should probably be addressed but then this is good to merge.

styles/styles.css
helms-charity: yes I've already authored the changes.

styles/styles.css
helms-charity: Yeah, we didn't have anything bigger than 48.

templates/mayura/mayura.css
helms-charity: you'll notice there is only 1 thing at 768, it is to hide the "old" sticky-cta container. We can remove this display:none when I merge in and author the new "cta-sticky" block.

blocks/footer/footer.css
helms-charity: Yeah it should be matching the footer on the Staging page much more closely now.

blocks/hero-heritage-cc/hero-heritage-cc.css
helms-charity: Ok I fixed most of those, the H1 will need to be re-authored as per our convo.

PR #448, opened 2026-01-26
Enhance footer accessibility by hiding decorative headings from scree…

### LINKED ISSUE #437: A11y score: footer h5, h6
We are losing points due to the H5 and H6 light text in the footer.

Image

Quick fix: In footer JS, add `aria-hidden="true"` to the first h5 and first h6 elements. (This attribute can be applied to elements to hide them from screen readers while still being visible to sighted users.)

sthirumal:
aria-hidden true only hides content from screen readers

Fix #437
Test URLs:
- After: https://437-footeraccess--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

After this fix the accessibility score improved from 89 to 91.

### DIFF
--- a/blocks/footer/footer.js
+++ b/blocks/footer/footer.js
@@ -18,6 +18,13 @@ export default async function decorate(block) {
 
   block.append(footer);
 
+  // Hide decorative headings from screen readers to resolve contrast audit
+  // Future suggestion: Replace with proper color contrast fix in CSS
+  const missionHeading = block.querySelector('h5[id^="on-a-mission"]');
+  const customerHeading = block.querySelector('h6[id^="customer-friendly"]');
+  if (missionHeading) missionHeading.setAttribute('aria-hidden', 'true');
+  if (customerHeading) customerHeading.setAttribute('aria-hidden', 'true');
+
   // Remove "only one open" behavior from footer accordions (allow multiple open)
   block.querySelectorAll('.accordion').forEach((accordion) => {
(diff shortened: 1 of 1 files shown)


sthirumal:
After adding the fix, the score improved from 89 to 91.

Having said that, following are the comments 

Aria-hidden only hides content from screen readers, not from sighted users.

Lighthouse's color contrast audit checks visual accessibility - it's for users with low vision who can see the screen but struggle with low contrast. Thats the reason when the accessibility is checked again this h5 and h6 will be highlighted again. 
To fully fix color contrast, you'll need to update the CSS in footer.css to change the text color from #bc3b45 to something with better contrast against the #9d1d27 background (e.g., white #ffffff).

Thank you,
Swapna

blocks/footer/footer.js
```diff
@@ -18,6 +18,13 @@ export default async function decorate(block) {
 
   block.append(footer);
 
+  // Hide decorative headings from screen readers to resolve contrast audit
+  // Future suggestion: Replace with proper color contrast fix in CSS
+  const missionHeading = block.querySelector('h5[id^="on-a-mission"]');
```
helms-charity: Using the ID from the heading is dangerous, because as soon as they update the text, the code won't work. So I think all you need to select for is `h5:first-of-type` and `h6:first-of-type `.

helms-charity (changes requested):

helms-charity (approved):

PR #461, opened 2026-01-27
only changed section breakpoint

### LINKED ISSUE #402: Update mobile size to 990px
Except for CARDS....

find and replace (both in CSS and JS) "900" with "990" for CSS and JS.
Take a look at what currently matches for "900": https://github.com/search?q=repo%3Aaemsites%2Fidfc+900&type=code

Also be cognizant of instances of "max-width" being 1 pixel less, such as:
https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L23

helms-charity:
I realized that the only thing causing a problem with the new 990px width breakpoint we added to the header were section backgrounds. And we're only using that feature in the "customer service" nav dropdown that I could find.

That "Download our App" section should look like this now from 0 - 989px wide:

image

I also fixed the width on this while I was in there.
image

Fix #402-990breakpoint

Test URLs:
- After: https://402-990breakpoint--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/header/header.css
+++ b/blocks/header/header.css
@@ -1002,7 +1002,7 @@ header nav .nav-sections .default-content-wrapper>ul>li .nav-h4-section-content 
 
 .cs-dropdown.open {
   visibility: visible;
-  padding: 40px var(--spacing-xxs) var(--spacing-xxs);
+  padding: 40px var(--spacing-s) var(--spacing-xs);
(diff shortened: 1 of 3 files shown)


zhummerz (approved):
Looks good. Approving.

PR #465, opened 2026-01-27
header dropdown font consistency fixes

### LINKED ISSUE #456: Font-related styles in header nav
Mobile and desktop, check both static font sizes/weights and also on hover.
Compare with when the header has the white background on https://stage.idfcfirst.bank.in/home-page-new .

bunting-adbe:
Fixed inconsistent font sizes and functions in the header nav dropdown. 

Fix #456 

Test URLs:
- After: https://456-nav-fonts--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Main:
image

Updates:
image


roperev (approved):
(edited: Charity noticed some breaking issues)

helms-charity (changes requested):
Those links are all supposed to have a hover color change. There's also a weird effect where the cursor is hand for the entire open nav panel, so you can't see what is a link and what isn't.

This is what the live site has:
Image

Here's where the cursor can be changed:
image

roperev:
> Those links are all supposed to have a hover color change. There's also a weird effect where the cursor is hand for the entire open nav panel, so you can't see what is a link and what isn't.
> 
> Image

I spoke too soon! 😨

bunting-adbe:
I noticed that also, but didn't address it. I will put that change in.

helms-charity (changes requested):
I see the hover effects now. Can you also try removing the 'cursor: pointer' at header.css:1780? Because the entire nav panel looks like it is clickable right now, but only the links are.
(or at least it was at line 1780 when I first reviewed)
image

bunting-adbe:
I have removed the mob-header file.
We do need the nav-pane.css file. It is styling the category nav panel in the header. 

With the nav-pane.css:
image

Without the nav-pane.css:
image

The fragment styling is still in place with both. We may be able to simplify what is actually in the nav-pane.css and incorporate that into the header.css, if we want.

helms-charity (approved):
looks good now

PR #479, opened 2026-01-28
Adjust modal background height for mobile width

### LINKED ISSUE #459: modal backdrop height in mobile
In mobile (widths >=990px for the header), there is no category nav, so the backdrop can start higher.

Image

techmaven:
Removes background image gap caused by hidden secondary nav menu at mobile width.

Fixes #459 

Test URLs:
- After: https://459-modal-backdrop-height--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/modal/modal.css
+++ b/blocks/modal/modal.css
@@ -981,5 +981,7 @@ body.modal-open {
     justify-self: center;
   }
   
-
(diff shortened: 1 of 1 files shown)


helms-charity (approved):
looks good

PR #481, opened 2026-01-28
Adjust fixed elements height calculation in anchor navigation

### LINKED ISSUE #473: Mayura anchor nav headerHeight
The current height being added to the position of the anchor-nav block is making it overlap behind the category nav (maybe by 6px).
See anchor-nav.js:26.

Image

compare to the live site
https://www.idfcfirst.bank.in/credit-card/metal-credit-card/mayura#

sthirumal:
Adjust height calculation in anchor navigation breadcrumb

Fix #473 
Test URLs:
- After: https://473-navhgt--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/anchor-nav/anchor-nav.js
+++ b/blocks/anchor-nav/anchor-nav.js
@@ -2,7 +2,7 @@
  * Calculate total height of fixed elements (header + category nav)
  */
 function getFixedElementsHeight(headerHeight) {
-  let totalHeight = headerHeight - 10;
+  let totalHeight = headerHeight - 6;
   const categoryNavWrapper = document.querySelector('.category-nav-wrapper');
   if (categoryNavWrapper?.offsetHeight > 0) {
     totalHeight += categoryNavWrapper.offsetHeight;


helms-charity (approved):
good job

sthirumal:
Here's the logic behind --sticky-top in the anchor-nav:

Purpose
The --sticky-top CSS custom property controls where the anchor-nav sticks when it becomes fixed to the viewport. It ensures the anchor-nav sits directly below any fixed elements (header, category nav) without overlapping them.

anchor-nav.css has the below.
top: var(--sticky-top, 54px)
When the .stuck class is added, the nav becomes position: fixed with top: var(--sticky-top, 54px). The 54px is a fallback if the JS hasn't set the value yet.

--sticky-top is calculated by adding the below :
Header height (minus 10px adjustment, likely for visual spacing)
Category nav height (if present on the page)

And the header height is calculated using below function:
function getFixedElementsHeight(headerHeight) {
  let totalHeight = headerHeight - 6;

Initially it was subtracting 10px; i readjusted to subtract 6px.

Hope the above helps.

Thank you,
Swapna

PR #492, opened 2026-01-29
mobile gutter size

helms-charity:
Mobile left and right padding was too much.
was 24px, now is 16.
Fix 455

Test URLs:
- After: https://455-mobilepadding--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/cards/cards.css
+++ b/blocks/cards/cards.css
@@ -175,15 +175,6 @@
 }
 
 /* === EARN REWARDS CARD VARIANT === */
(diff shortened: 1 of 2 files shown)


blocks/cards/cards.css
```diff
@@ -175,15 +175,6 @@
 }
 
 /* === EARN REWARDS CARD VARIANT === */
```
amarghioali, line 177: We should probably remove this comment too

amarghioali (approved):
LGTM

PR #507, opened 2026-01-31
Add H1 to main then hide with CSS

### LINKED ISSUE #378: Add H1 to Main and hide it
First, change the JSON field name from "h1Title" to "h1-title". 
Then push that change/get it merged so you can author a new h1-title and you should see this in the page metadata after publishing.

Use JS to set that value as an H1 element prepended to "main".
- prepend the H1 tag and it's value to Main so it is the first element, before any other divs.
- Now hide it, like:
```
main > h1:first-child {
  display: none;
}
```

techmaven:
To address Google complaint of missing H1.

Fix #378 

Test URLs:
- After: https://378-make-hide-h1--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -811,6 +811,12 @@ export function decorateMain(main) {
   decorateBlocks(main);
   buildEmbedBlocks(main);
   decorateLinkedPictures(main);
+  const h1Title = getMetadata('h1-title');
+  if (h1Title) {
+    const h1 = document.createElement('h1');
+    h1.textContent = h1Title;
+    main.insertBefore(h1, main.firstChild);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


helms-charity (changes requested):
I found an interesting bug. The H1 is added to the page entrance modal. There must be some code in the modal.js bringing in the page H1? See if you can fix that with your PR.

Image

Image

Image

helms-charity (commented):
That didn't fix it, I still see the H1 in the modals on both pages...

techmaven:
@helms-charity I moved creation of h1 outside of decorateMain(), appears to me to resolve the modal issue.

helms-charity (approved):
Putting it in Eager makes perfect sense. Looks good.

PR #522, opened 2026-02-03
fix: show bell sections in Universal Editor

### LINKED ISSUE #521: Bell notification sections hidden in Universal Editor preview panel
**Problem**
After merging the CLS prevention fix for bell notification sections, content is now hidden in the Universal Editor preview panel. Authors can see the Cards block in the layers panel but cannot preview or edit the content visually.

**Root Cause**
The CSS fix that prevents CLS by hiding bell sections uses a global selector that applies in all contexts, including Universal Editor:
```

.section.cards-container[id*="bell"] {
  display: none !important;
}

```

amarghioali:
Fix #521

The CLS prevention CSS was hiding bell notification sections globally, including in the Universal Editor where authors need to preview and edit content.

**Changes**:

- Scope bell hiding rules to exclude Universal Editor using main:not([data-aue-resource]) selector
- Scope bell hiding rules to exclude framework pages using body:not(.is-framework-page) selector
- Add explicit show rules for Universal Editor and framework pages
- Remove overly generic :has(.default-content .icon) selector that could hide unrelated cards-container sections
- Remove broad [id="outline"] selector for safety

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/rupay-credit-card1
- After:  https://521-ue-fix--idfc--aemsites.aem.page/drafts/rupay-credit-card1

### DIFF
--- a/blocks/category-nav/category-nav.css
+++ b/blocks/category-nav/category-nav.css
@@ -47,21 +47,39 @@ body.is-framework-page .category-nav.block > div > div {
    Bell sections are Cards blocks with icons that get moved to header nav.
    Hiding them immediately prevents Cumulative Layout Shift.
    
-   TODO: Refactor to make bell sections proper category-nav blocks
-         See GitHub issue for proper architecture implementation
+   NOTE: Exceptions:
+     - Universal Editor (main[data-aue-resource]) - content must be visible for editing
+     - Framework pages (body.is-framework-page) - content must be visible for preview/editing
    ================================================================= */
 
-/* Hide bell notification sections immediately to prevent CLS */
-.section.cards-container[id*="bell"],
-.section[data-category-id*="bell"],
-.section#bell-outline,
-.section[id*="outline"].cards-container {
+/* Hide bell notification sections on normal pages to prevent CLS */
+body:not(.is-framework-page) main:not([data-aue-resource]) .section.cards-container[id*="bell"],
+body:not(.is-framework-page) main:not([data-aue-resource]) .section[data-category-id*="bell"],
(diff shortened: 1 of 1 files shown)


blocks/category-nav/category-nav.css
```diff
@@ -47,21 +47,39 @@ body.is-framework-page .category-nav.block > div > div {
...
-.section.cards-container[id*="bell"],
-.section[data-category-id*="bell"],
-.section#bell-outline,
-.section[id*="outline"].cards-container {
+/* Hide bell notification sections on normal pages to prevent CLS */
+body:not(.is-framework-page) main:not([data-aue-resource]) .section.cards-container[id*="bell"],
+body:not(.is-framework-page) main:not([data-aue-resource]) .section[data-category-id*="bell"],
+body:not(.is-framework-page) main:not([data-aue-resource]) .section#bell-outline {
   display: none !important;
+  visibility: hidden !important;
```
bunting-adbe, line 60: Are all of these `!important` tags necessary? If the selector is specific enough these are not needed and typically we try to avoid.

bunting-adbe (approved):
Can you double check if you need the `!important` tags on all of the selectors? 
I do see the bell section in the UE now, but I wasn't able to quickly determine the way to edit it. That may be something to more clearly demonstrate. I approve as it does correct the issue at hand.

blocks/category-nav/category-nav.css
helms-charity: that's a good callout. I think I can be talked into it here because no matter what other styles others may add, we never want this to display in mobile due to the bad CLS score, so I think we can allow it. (also I want to merge so our cards are no longer missing in the UE).

helms-charity (approved):

PR #532, opened 2026-02-04
Refactor swiper pagination handle for improved drag functionality and…

### LINKED ISSUE #518: Mayura card scroller icons
New scrollbar with icons was added to Mayura cards via https://github.com/aemsites/idfc/pull/506/changes .

However, we noticed that if you start your browser in desktop size and shrink down to mobile size,
The scroll handle icon doesn't appear in mobile view for Rewards Points or Explore other cards cards that started out with no scrollbar showing:

Image

Image

sthirumal:
Added dynamic handle creation and ensured handle persistence during Swiper re-renders. Updated event listeners for drag and resize events.

Fix #518

Test URLs:
- After: https://518-scicon--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/cards/cards.js
+++ b/blocks/cards/cards.js
@@ -1123,46 +1123,17 @@ export default async function decorate(block) {
     if (isMayuraTemplate) {
       const paginationEl = block.querySelector('.swiper-pagination');
       if (paginationEl) {
-        const handleIndicator = document.createElement('div');
-        handleIndicator.className = 'swiper-pagination-handle';
-        // Using scrollbar-handle.svg as the indicator
-        handleIndicator.innerHTML = '<img src="/icons/scrollbar-handle.svg" alt="" />';
-        paginationEl.appendChild(handleIndicator);
-
-        // Update handle indicator position based on swiper progress
-        const updateHandlePosition = () => {
-          const { progress } = swiper;
-          // Clamp progress between 0 and 1
-          const clampedProgress = Math.max(0, Math.min(1, progress));
-          handleIndicator.style.left = `${clampedProgress * 100}%`;
-        };
-
-        // Initial position update
-        updateHandlePosition();
-
-        // Update handle position on slide change and progress events
-        swiper.on('progress', updateHandlePosition);
-        swiper.on('slideChange', updateHandlePosition);
-
-        // Make progressbar clickable to navigate to slides
-        paginationEl.style.cursor = 'pointer';
-        paginationEl.addEventListener('click', (e) => {
-          // Ignore clicks on the handle itself (handled by drag)
-          if (e.target.closest('.swiper-pagination-handle')) return;
-          const rect = paginationEl.getBoundingClientRect();
-          const clickX = e.clientX - rect.left;
-          const percentage = clickX / rect.width;
-          const targetSlide = Math.round(percentage * (slideCount - 1));
-          swiper.slideTo(targetSlide);
-        });
-
-        // Make handle indicator draggable for slide navigation
+        // Drag state (shared across handle instances)
         let isDragging = false;
 
+        // Drag event handlers
         const handleDragStart = (e) => {
           isDragging = true;
-          handleIndicator.style.transition = 'none';
-          handleIndicator.style.cursor = 'grabbing';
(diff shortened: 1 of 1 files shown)


sthirumal:
Root Cause Analysis
The issue is that when Swiper transitions between locked state (all slides visible, swiper-pagination-lock applied) and unlocked state (pagination needed), Swiper internally updates/re-renders the pagination element. During this process, it clears the pagination content, which removes any custom children like the swiper-pagination-handle element.

Sequence of events causing the bug:

Load at wide viewport (1400px): Swiper initializes → All 2 slides fit → swiper-pagination-lock is added → Handle IS created and appended
Resize to mobile (550px): Swiper detects fewer slides fit → Removes swiper-pagination-lock → Swiper re-renders pagination internally, clearing the handle
Refresh at mobile: Fresh initialization → Handle is created → Works correctly
The Fix (lines 1122-1230)

createHandleIndicator() function (lines 1171-1179) - Creates a new handle element with proper event listeners attached
ensureHandleExists() function (lines 1182-1190) - Checks if handle exists in the pagination; if missing, recreates it
updateHandlePosition() now calls ensureHandleExists() (line 1199) - Every time the position is updated, it first ensures the handle exists
Added resize and breakpoint event listeners (lines 1216-1217):
swiper.on('resize', updateHandlePosition); swiper.on('breakpoint', updateHandlePosition);
These events fire when viewport changes and Swiper updates its internal state. The updateHandlePosition function will detect the missing handle and recreate it.
Drag handlers query the handle dynamically (lines 1132, 1150, 1163) - Instead of holding a stale reference, the drag handlers now query for the current handle element each time

blocks/cards/cards.js
```diff
@@ -1123,46 +1123,17 @@ export default async function decorate(block) {
     if (isMayuraTemplate) {
       const paginationEl = block.querySelector('.swiper-pagination');
```
helms-charity, line 1124: This is OK for this PR. ⭐ 
 In another ticket, we'll be allowing authors to choose if they want dot nav or draggable nav. So we'll want to separate the draggable pagination function from the specific Mayura decoration of it in the future.

helms-charity (changes requested):
I just noticed we lost the smooth card transition animation with the new slider. (you can see it working as expected when you click on a dot in the Rupay dot navigation, and on the customer Mayura page). 

image

sthirumal:
@helms-charity As per your suggestion when I started working to remove the js changes and only using the css to display and hide swiper-pagination-handle based on swiper-pagination-lock.

Looks like the css doesnt work due to the following reason :

Looking at your DOM examples, the problem is not about visibility - the swiper-pagination-handle element is being completely removed from the DOM when Swiper transitions from locked to unlocked state.

The Root Cause
When Swiper detects a breakpoint/resize change that toggles between locked/unlocked states, it internally re-renders the pagination element. During this re-render, Swiper clears the pagination's children (keeping only its own swiper-pagination-progressbar-fill), which removes your custom swiper-pagination-handle element.

The Only Solution
The fix must be in JavaScript to detect when the handle is missing and recreate it.

sthirumal:
Comment on code changes and its working

Key Features
-> All comments start with "swapna" - Easy to identify the fix in the codebase

->  Performance Optimized:
     -  Uses a single global state object (window.swiperHandleDragState) instead of per-instance variables
     - Only 4 document-level listeners total (mousemove, touchmove, mouseup, touchend) regardless of how many swipers exist on the page
       - Early returns in event handlers when not dragging (minimal CPU overhead)

-> Only 1 Instance of Listeners Per Page:
      - Global listeners are registered only once using the if (!window.swiperHandleDragState) check (line 1127)
      - Multiple swipers share the same global drag handlers
      - The active swiper is tracked via state.activeSwiper so the correct instance handles the drag


The way its working, when viewport changes from 1400px → 550px:
1. Swiper fires resize and/or breakpoint events 
2. updateHandlePosition() is called
3. ensureHandleExists() detects the handle is missing
4. A new handle is created with proper event listeners
5. The handle appears correctly without needing a page refresh

helms-charity (approved):

PR #538, opened 2026-02-04
Initial - hotspot block 

iustinp:
some json and some js

Fix # work on 534

Test URLs:
- After: https://534-hotspot-block--idfc--aemsites.aem.page/


blocks/hotspot/_hotspot.json
```diff
@@ -0,0 +1,100 @@
...
+                "valueType": "string",
+                "name": "image",
+                "label": "Image",
+                "multi": false
+            },
+            {
+                "component": "text",
+                "name": "id",
+                "label": "ID",
+                "description": "Adds an ID value to the section"
```
helms-charity: Adds an ID value to the BLOCK (so that other hotspot blocks can link to it)

helms-charity (changes requested):
just a comment

helms-charity (approved):
I think you can leave the X and Y text coordinate off for now.

PR #541, opened 2026-02-04
Center Span Icons

### LINKED ISSUE #394: Text elements + span vertical alignments
Whenever we have a text element and a span.icon element next to each other, they should be vertically aligned center.

These sparkles are too high:
Image

I did something like this specific to buttons that have an icon after them, so maybe we can expand on this to include other cases.

Image

Be careful not to break other places that have icons + text, but DO remove any individual block or section code that will become redundant with this new change.

bunting-adbe:
Center the sparkle icons horizontally with the included text. 

Fix #394 

Test URLs:
- After: https://394-spakleAlignment--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Main:
image

Branch:
image

### DIFF
--- a/styles/styles.css
+++ b/styles/styles.css
@@ -516,6 +516,14 @@ a.button:has(span.icon) {
   justify-content: center;
 }
 
+/* Align icons within headings (similar to button icon alignment above) */
+h3:has(span.icon) {
+  display: inline-flex;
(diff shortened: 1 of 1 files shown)


zhummerz (approved):
Looks good and sparkly.

PR #542, opened 2026-02-05
Section Title UE changes

### LINKED ISSUE #535: Block: section-title
### Problem:
We need a way for authors to set the size and alignment of their titles/subtitles without breaking semantic HTML best practices. However, we can't update the OOTB Title with resource type "core/franklin/components/title/v1/title".

Image

### Solution:
We can create a new block that functions similarly to the OOTB Title with resource type "core/franklin/components/block/v1/block" instead.
This block should probably not have any top/bottom margin or padding, since the individual heading elements will.

### Testing:
Please test with 2 of these blocks in a row, one with an H2 and one an H3 for example. 

Please test this also with :icon-notation: in the text field for cases like this:

Image

Don't forget to allow 'section-title' as an allowed block under _section.json's filters.

### Example
Some sample JSON config for this idea is below:

```
      "fields": [
        {
          "component": "select",
          "name": "classes",
          "label": "Alignment",
          "value": "",
          "options": [
            {
              "name": "Default",
              "value": ""
            },
            {
              "name": "Center",
              "value": "center"
            },
        {
          "component": "select",
          "name": "sizes",
          "label": "Text size", // in the JS, set an inline style for the element created to whatever the value is.
          "value": "",
          "options": [
            {
              "name": "Default",
              "value": ""
            },
            {
              "name": "XXL",
              "value": "var(heading-font-size-xxl)" //  we need to test that parentheses will write back to the JCR and get converted to .md
            },
         // (continue with XL through XS...)
 
        },
        {
          "component": "text",
          "name": "title",
          "label": "Title"
        },
        {
          "component": "select",
          "name": "titleType",
          "label": "Title Type",
          "options": [
            {
              "name": "h1",
              "value": "h1"
            },
            {
              "name": "h2",
              "value": "h2"
            },
            {
              "name": "h3",
              "value": "h3"
            },
            {
              "name": "h4",
              "value": "h4"
            },
            {
              "name": "h5",
              "value": "h5"
            },
            {
              "name": "h6",
              "value": "h6"
            },
            {
              "name": "p",
              "value": "p"
            }
          ]
```

bunting-adbe:
Initial UE changes for new Section-Title block. 

Fix #535 

Test URLs:
- After: https://535-sectionTitle--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/section-title/_section-title.json
+++ b/blocks/section-title/_section-title.json
@@ -67,6 +67,43 @@
         {
           "component": "select",
           "name": "classes",
+          "label": "Text Size",
+          "value": "",
+          "valueType": "string",
+          "options": [
+            {
+              "name": "Default (based on heading type)",
+              "value": ""
+            },
+            {
+              "name": "XXL (45px/48px)",
+              "value": "size-xxl"
+            },
+            {
+              "name": "XL (36px/42px)",
+              "value": "size-xl"
+            },
+            {
+              "name": "L (32px/36px)",
+              "value": "size-l"
+            },
+            {
+              "name": "M (28px/32px)",
+              "value": "size-m"
+            },
+            {
+              "name": "S (24px/28px)",
+              "value": "size-s"
+            },
+            {
+              "name": "XS (22px/24px)",
+              "value": "size-xs"
+            }
+          ]
+        },
+        {
+          "component": "select",
+          "name": "classes",
           "label": "Alignment",
           "value": "",
           "valueType": "string",
--- a/blocks/section-title/_section-title.json
+++ b/blocks/section-title/_section-title.json
@@ -1,0 +1,92 @@
+{
+  "definitions": [
+    {
+      "title": "Section Title",
+      "id": "section-title",
+      "plugins": {
+        "xwalk": {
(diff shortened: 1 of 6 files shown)


helms-charity (approved):
I guess you didn't want to try the size yet.

bunting-adbe:
Shoot, i forgot that. I just added it. Did I miss anything else that we talked about?

blocks/section-title/_section-title.json
```diff
@@ -0,0 +1,129 @@
...
+          "value": "",
+          "valueType": "string",
+          "options": [
+            {
+              "name": "Default (based on heading type)",
+              "value": ""
+            },
+            {
+              "name": "XXL (45px/48px)",
+              "value": "size-xxl"
```
helms-charity, line 80: I struggle with putting the pixel values here, because what if they change the variable's values? I was figuring they were eyeballing what it looks like anyway when they author. I think it's best to leave that off.

helms-charity (approved):
I left a comment about using the px value in the UE dropdown names, but I'm approving this anyway.

PR #547, opened 2026-02-05
fix(category-nav): Remove bell section from DOM instead of hiding (#546)

### LINKED ISSUE #546: Fix: Remove bell notification section from DOM instead of hiding it
## Issue

https://main--idfc--aemsites.aem.page/credit-card/rupay-credit-card

The bell notification menu item (category-nav-bulletin-notification) is functioning correctly in the header, but the original section containing the Cards block remains in the DOM at the beginning of the <main> element. This section is currently only hidden with display: none, which leaves unnecessary DOM pollution.

Image

amarghioali:
Fix #546 

Test URLs:
- After: https://546-bell-section--idfc--aemsites.aem.page/credit-card/rupay-credit-card


helms-charity (approved):
that's better

PR #549, opened 2026-02-05
Section Title UE correction

### LINKED ISSUE #535: Block: section-title

bunting-adbe:
Needed to update the name on 2 component parts. They were not showing up in the UE.

Fix #535 

Test URLs:
- After: https://535-sectionTitle--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/section-title/_section-title.json
+++ b/blocks/section-title/_section-title.json
@@ -66,7 +66,7 @@
         },
         {
           "component": "select",
-          "name": "classes",
+          "name": "textSize",
           "label": "Text Size",
           "value": "",
           "valueType": "string",
@@ -103,7 +103,7 @@
         },
         {
           "component": "select",
-          "name": "classes",
+          "name": "alignment",
           "label": "Alignment",
           "value": "",
           "valueType": "string",
--- a/component-models.json
+++ b/component-models.json
@@ -1619,7 +1619,7 @@
       },
       {
         "component": "select",
-        "name": "classes",
+        "name": "textSize",
         "label": "Text Size",
         "value": "",
         "valueType": "string",
@@ -1656,7 +1656,7 @@
       },
       {
         "component": "select",
-        "name": "classes",
+        "name": "alignment",
         "label": "Alignment",
         "value": "",
         "valueType": "string",


zhummerz (approved):
approved

PR #557, opened 2026-02-05
UE updates for testimonial cards

### LINKED ISSUE #133: Testimonial cards enhancements

Card item level config --

bunting-adbe:
UE updates for testimonial cards so that they can be authored and added to schema. This is UE changes only for now. I authored a new card using the changes on my branch, but need them to added to the bus to properly test additional code change needs. 

Fix #133 

Test URLs:
- After: https://133-testimonialEnhancement--idfc--aemsites.aem.page/credit-card/rupay-credit-card

Just adding UE elements now. There is some more testing needed with the JS and CSS once we get these in place. 
Added the testimonial card input fields at the card level in an expandable container. 
image

### DIFF
--- a/blocks/cards/_cards.json
+++ b/blocks/cards/_cards.json
@@ -219,6 +219,64 @@
           "label": "Modal Content (optional)",
           "description": "For simple modals, author content here. For complex modals, use Card Link pointing to /modals/ path instead.",
           "valueType": "string"
+        },
+        {
+          "component": "container",
+          "name": "testimonialDetails",
+          "label": "Testimonial (optional)",
+          "collapsible": true,
+          "description": "Only used when block Style is Testimonial. For schema and display.",
+          "fields": [
+            {
+              "component": "select",
+              "name": "quoteIcon",
+              "label": "Quote Icon",
+              "valueType": "string",
+              "description": "Icon shown at top of this testimonial card.",
+              "options": [
+                {
+                  "name": "None",
+                  "value": ""
+                },
+                {
+                  "name": "Inverted commas",
+                  "value": "inverted-commas"
+                }
+              ]
+            },
+            {
+              "component": "text",
+              "name": "personName",
+              "label": "Person Name",
+              "valueType": "string",
+              "description": "Author or reviewer name"
+            },
+            {
+              "component": "text",
+              "name": "productName",
+              "label": "Product Name",
+              "valueType": "string",
+              "description": "Product being reviewed (e.g. card name)"
+            },
+            {
+              "component": "number",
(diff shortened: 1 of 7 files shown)


blocks/cards/_cards.json
```diff
@@ -219,6 +219,64 @@
           "label": "Modal Content (optional)",
           "description": "For simple modals, author content here. For complex modals, use Card Link pointing to /modals/ path instead.",
           "valueType": "string"
+        },
+        {
+          "component": "container",
+          "name": "testimonialDetails",
+          "label": "Testimonial (optional)",
+          "collapsible": true,
+          "description": "Only used when block Style is Testimonial. For schema and display.",
```
helms-charity: Check this out -- you can 'test' for if the style is testimonial, only then show this container.
https://github.com/aemsites/world-bank/blob/main/models/_section.json#L104-L111

blocks/cards/_cards.json
```diff
@@ -219,6 +219,64 @@
...
+          "fields": [
+            {
+              "component": "select",
+              "name": "quoteIcon",
```
helms-charity: I think it is safest to use all small letters with a hyphen instead of camelcase, because either the JCR or the markdown process changes these in certain conditions. (same for personName, productName, authorRating, reviewDate)

blocks/cards/_cards.json
```diff
@@ -219,6 +219,64 @@
...
+                "numberMin": 1,
+                "numberMax": 5
+              }
+            },
+            {
+              "component": "text",
+              "name": "reviewDate",
+              "label": "Date",
+              "valueType": "string",
+              "description": "Review date for Schema.org. Use YYYY-MM-DD (e.g. 2025-03-26) or readable (e.g. March 26, 2025)."
```
helms-charity: hmm, let's tell them to use only YYYY-MM-DD (because we're requiring the same format in page properties for a custom date there too), and your script can translate it to Month day, year.

helms-charity (changes requested):

helms-charity (approved):

blocks/cards/_cards.json
bunting-adbe: Found this doesn't work for our use case.

blocks/cards/_cards.json
bunting-adbe: changed this before we decided to split the blocks. Cursor went with camel case because there were several other examples of it in the cards file.

blocks/cards/_cards.json
bunting-adbe: fixed this

Issue #536, opened 2026-02-04
Mayura "The Concept" modal animation


helms-charity:
Compare with the live page.

### Starting/ending modal animation
- The live site animation does a slide up/down when you click on "The Concept" which we want to implement too.
- Our animation currently does a fade in/out (I think per https://github.com/aemsites/idfc/blob/main/blocks/hero-heritage-cc/hero-heritage-cc.css#L327 )

### Swapping each card hotspot section
<img width="1069" height="762" alt="Image" src="https://github.com/user-attachments/assets/e0530ee4-2838-48cf-b5db-1d8cf1ec1de3" />

techmaven:
Per @helms-charity part two is being handled separately by @iustinp

PR #558, opened 2026-02-06
Hotspot block functionality & styling

### LINKED ISSUE #534: Block: hotspot, with hotspot-items
Since the hot spot concept is used on more than one page for IDFC (such as https://www.idfcfirst.bank.in/credit-card/metal-credit-card/ashva but there are probably more), we should make this feature more authorable. 

Currently we are using 1 section with 1 columns block per image use, and we only have one hotspot per image.
The hotspot block should eventually allow more hotspots one a single image by allowing the author to add each hotspot-item per hotspot block, but we can start with assuming only 1 hotspot-item for this ticket.

See this for inspiration
https://main--franklin-assets-selector--hlxsites.aem.live/hotspot-block/hotspot?view-doc-source=true
https://github.com/hlxsites/franklin-assets-selector/blob/main/blocks/hotspot/hotspot.js

I think maybe using the Carousel boilerplate block might be a good model to follow? https://github.com/aemdemos/sta-boilerplate/tree/main/blocks/carousel

In the JS, we need functionality where a hotspot-item can be linked to another hotspot block’s ID to load that. Perhaps   “if the richText has only a button, and that a href starts with “#”, then just link to that….”?

Image

### Assumptions
So to use 2 English sayings, we don't need to "boil the ocean" by creating more than 1 hotspot-item or having text vs. video (like that other hotspot block has), but also we shouldn't "paint ourselves into a corner" by making our code too specific on our current use-case.

You can still assume we'll have the blocks on a fragment page, though it would be great if instead of having 3 separate sections, we put 3 hotspot blocks in the same section.

iustinp:
To use a Romanian expression, I have "taken the bread from @techmaven's mouth", cause i've also done the animations (#536) :) 

One thing i'm not sure of: When u're on the 1st card and you click Go Back, I made it hide everything (the whole hotspot section) but i'm not sure it's necessarily helpful when you want to integrate this. I guess it can be tweaked later. 

Fix #534

Test URLs:
- After: https://534-hotspot-block--idfc--aemsites.aem.page/drafts/test

### DIFF
--- a/blocks/hotspot/hotspot.css
+++ b/blocks/hotspot/hotspot.css
@@ -55,6 +55,7 @@
   line-height: 1.6;
   position: relative;
   padding-right: 30px;
+  transition: padding-top 0.5s ease-in-out;
 }
 
 .hotspot-panel-item {
--- a/blocks/hotspot/hotspot.css
+++ b/blocks/hotspot/hotspot.css
@@ -12,7 +12,19 @@
   width: 100%;
   min-height: 400px;
   position: relative;
-  transition: opacity 0.3s ease;
+  transition: opacity 0.2s ease;
+}
+
+/* SVG connector line styles (line drawn via JS) */
+.hotspot-connector-svg {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  pointer-events: none;
+  overflow: visible;
+  z-index: 1;
 }
 
 /* Fade animation for block transitions */
@@ -33,7 +45,7 @@
   flex: 0 0 50%;
   display: flex;
   align-items: flex-start;
-  justify-content: flex-end;
+  justify-content: flex-start;
   padding: 40px;
   position: relative;
   opacity: 0;
@@ -54,7 +66,7 @@
   font-size: 18px;
   line-height: 1.6;
   position: relative;
-  padding-right: 30px;
+  padding-right: 0;
   transition: padding-top 0.5s ease-in-out;
 }
 
@@ -68,10 +80,9 @@
 
 .hotspot-panel-item p {
   margin: 0;
-  background: var(--Primary, linear-gradient(308.18deg, #110140 4.16%, #1E2A8E 44.91%, #010740 88.8%));
+  background: var(--primary, linear-gradient(308.18deg, #110140 4.16%, #1E2A8E 44.91%, #010740 88.8%));
   background-clip: text;
-  -webkit-background-clip: text;
-  -webkit-text-fill-color: rgba(0, 0, 0, 0);
+  -webkit-text-fill-color: rgb(0 0 0 / 0%);
   font-size: 16px;
   font-weight: 500;
   line-height: normal;
@@ -79,6 +90,16 @@
 
 .hotspot-panel-item p:last-child {
   margin-bottom: 0;
+}
+
+.hotspot-panel-item > p > a.button {
+  color: rgb(0 0 0);
+  font-size: 16px;
+  font-weight: 500;
+  line-height: normal;
+  text-decoration-line: underline;
+  margin-top: 8px;
+  padding: 0;
 }
 
 .hotspot-panel-item em {
@@ -100,18 +121,6 @@
 
 .hotspot-go-back a:hover {
   text-decoration: none;
-}
-
-/* SVG connector line styles (line drawn via JS) */
-.hotspot-connector-svg {
-  position: absolute;
-  top: 0;
-  left: 0;
-  width: 100%;
-  height: 100%;
-  pointer-events: none;
-  overflow: visible;
-  z-index: 1;
 }
 
 .hotspot-connector-svg line {
@@ -204,7 +213,7 @@
 }
 
 /* Responsive adjustments */
-@media (max-width: 1024px) {
+@media (width <= 1024px) {
   .hotspot-container {
     min-height: 350px;
   }
@@ -223,7 +232,7 @@
   }
 }
 
-@media (max-width: 768px) {
+@media (width <= 768px) {
   .hotspot-container {
     flex-direction: column;
     min-height: auto;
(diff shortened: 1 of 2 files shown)


blocks/hotspot/hotspot.css
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/* Two-column layout */
+.hotspot-container {
+  display: flex;
+  align-items: stretch;
+  width: 100%;
+  min-height: 400px;
+  position: relative;
+  transition: opacity 0.3s ease;
```
helms-charity: Can we change the timing to 0.1s? The live site doesn't have a fade in and out, so although I like having it, let's make it really quick.

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+  // Row 2: Hotspot Text (block-level text for left panel)
+  // Check if this row is hotspot-text (1 cell) or a hotspot item (3 cells with coordinates)
+  if (rows.length > metadataRowCount) {
+    const potentialTextRow = rows[metadataRowCount];
+    const cells = Array.from(potentialTextRow.children);
+    
+    // If row has 1 cell with content (not an image, not starting with #), it's hotspot-text
+    if (cells.length === 1) {
+      const textCell = cells[0];
+      if (!textCell.querySelector('picture')) {
```
helms-charity: why limit what can be in here? I can definitely see them having some icon in with their text.

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+  const rows = Array.from(block.children);
+
+  // Extract block-level fields
+  let metadataRowCount = 0;
+  let imageElement = null;
+  let blockId = '';
+  let hotspotText = '';
+
+  // Row 0: Image
+  if (rows.length > 0) {
```
helms-charity, line 590: this will always be true when the block is added (as you have it set up, there will always be rows as soon as you add the block)

Image

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+      if (blockId) {
+        metadataRowCount = 2;
+      }
+    }
+  }
+
+  // Row 2: Hotspot Text (block-level text for left panel)
+  // Check if this row is hotspot-text (1 cell) or a hotspot item (3 cells with coordinates)
+  if (rows.length > metadataRowCount) {
+    const potentialTextRow = rows[metadataRowCount];
```
helms-charity: You don't need to check for this, if you've added a hotspot-item, then all the cells in the row are present. So you can assume rows[2] will be that block-level text.

Image

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+        const cellText = textCell.textContent?.trim() || '';
+        if (cellText && !cellText.startsWith('#')) {
+          hotspotText = textCell.innerHTML;
+          metadataRowCount += 1;
+        }
+      }
+    }
+    // If row has 2 cells (first is text, second could be interpreted as a single coordinate row)
+    // This handles cases where the CMS might structure it differently
+    else if (cells.length === 2) {
```
helms-charity: there won't be a single coordinate row, so you can simplify this IF as well.

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+  showInitialPanel(container, blockId);
+
+  // Measure block height BEFORE hiding (for consistent section height)
+  const blockHeight = block.offsetHeight;
+  if (blockHeight > maxHotspotHeight) {
+    maxHotspotHeight = blockHeight;
+    // Update section min-height to accommodate the tallest block
+    const sectionWrapper = block.closest('.section');
+    if (sectionWrapper) {
+      sectionWrapper.style.minHeight = `${maxHotspotHeight}px`;
```
helms-charity: I like that you are adding this minHeight

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+    }
+  }
+
+  // Hide non-first blocks (only show the first hotspot block initially)
+  if (blockId !== firstBlockId) {
+    block.style.display = 'none';
+    block.setAttribute('aria-hidden', 'true');
```
helms-charity, line 855: You are hiding ALL other blocks on the page. Let's limit this to blocks within the same section for now?

blocks/hotspot/hotspot.js
```diff
@@ -1,52 +1,876 @@
...
+
+    // Store position for connecting line
+    hotspot.dataset.x = group.x;
+    hotspot.dataset.y = group.y;
+
+    moveInstrumentation(group.row, hotspot);
+
+    // Build panel content from block-level hotspot-text
+    const panelHTML = hotspotText
+      ? `<div class="hotspot-panel-item">${hotspotText}</div><div class="hotspot-go-back"><a href="#" data-go-back="${currentBlockId}">Go back</a></div>`
```
helms-charity: I don't want to add this "hotspot-go-back" div. I think we should allow the author to add a link to #whatever-block-id manually if they want a "go back" link.

helms-charity (changes requested):
I have quite a few comments this time.

blocks/hotspot/hotspot.js
iustinp: Refactored this so that authors can add #id-of-some-other-hotspot-block in the Go back link, which is added together with the hotspot-text. Now it jumps 3-2-1-gone.

blocks/hotspot/hotspot.js
iustinp: Fixed, together with some other behaviours that this brought to light, when you have multiple sections with this block on the page.

iustinp:
I've simplified the whole detection of fields for the above comments.

helms-charity (approved):
This looks pretty good to me now, please author the /drafts/mayura page however needed after merging though so we can see how it will compare with all the regular content.

PR #565, opened 2026-02-06
add testimonial card to section menu part 2

### LINKED ISSUE #133: Testimonial cards enhancements

bunting-adbe:
I misspelled my block id. Please help me. 

Fix #133 

Test URLs:
- After: https://133-testimonialEnhancement--idfc--aemsites.aem.page/


techmaven (approved):
Looks good.

PR #579, opened 2026-02-08
Multi section for accordions and tabs

### LINKED ISSUE #563: tabs-section
We need to replace all those tabs with fragments with a proper solution, see https://github.com/aemsites/ams-xwalk/wiki/Tabs-or-accordion-container .

helms-charity:
### New "multi-tab" section for content and blocks to be view inside tabs or accordions.

This takes multi-sections with the same name and gathers them together as tab or accordion content. The background options are reduced here but the styles that are used on SECTION ONE are used as the multi-section styles. 

The UE json was merged already, you can see it at https://github.com/aemsites/idfc/blob/main/models/_multi-section.json .

Once this is approved, the tabbed cards and accordions on Mayura will need to be reauthored.
There will no longer be a standalone tabs block that authors can use, but you can use standalone accordions.

Getting great scores.

To do: I wasn't really able to optimize what is going on in the UE yet, it seems I can't test my change to editor-support without merging.

Fix #563 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/tab-section
- After: https://563-tabsection--idfc--aemsites.aem.page/library/tab-section

### DIFF
--- a/blocks/accordion/accordion.css
+++ b/blocks/accordion/accordion.css
@@ -30,6 +30,10 @@
   font-weight: 600;
 }
 
+.multi .accordion details summary {
+  padding: 20px 30px 20px var(--spacing-s);
+}
+
 .accordion details summary::after {
   content: '';
   position: absolute;
--- a/blocks/accordion/accordion.js
+++ b/blocks/accordion/accordion.js
@@ -1,6 +1,7 @@
-import { moveInstrumentation } from '../../scripts/scripts.js';
+import { moveInstrumentation, getBlockId } from '../../scripts/scripts.js';
 
 export default function decorate(block) {
+  block.id = getBlockId('accordion');
   let noSchema = false;
   let ctaUrl = null;
   let ctaText = null;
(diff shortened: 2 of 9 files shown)


iustinp (commented):
In general I don't see any issues, it seems to work as intended, but here are just some observations:
- Existing styling for blocks included in these might need to be adapted maybe, because of the extra levels of depth, or because of the fact that these too are also called 'section' like the normal ones, and it might confuse some CSS?
- In the json:  "label": "Spacing top and bottom of section (xs, s, m, l, xl, xxl)". Maybe specify what the default will be if they go off rails?
- Maybe an option to have the width of the multi-section as wide as the whole page?
- I can't copy from the same page though. It seems you cannot copy blocks from sections to multi-sections and viceversa. Getting: "Item cannot be pasted into restricted containers."

Will continue looking tomorrow if needed.

techmaven (approved):
I don't find any issues. I reviewed the library page in UE and features appear to work as intended.

helms-charity:
@iustinp here are my replies:

> 
> * Existing styling for blocks included in these might need to be adapted maybe, because of the extra levels of depth, or because of the fact that these too are also called 'section' like the normal ones, and it might confuse some CSS?

They are indeed sections. A section is created with the tabs (or accordion) block, so I didn't have to update a lot of CSS.

> * In the json:  "label": "Spacing top and bottom of section (xs, s, m, l, xl, xxl)". Maybe specify what the default will be if they go off rails?

The default is none/0px.

> * Maybe an option to have the width of the multi-section as wide as the whole page?

There isn't a use-case for this now, so it using the same as the rest of the page content. Unless I'm misunderstanding?

> * I can't copy from the same page though. It seems you cannot copy blocks from sections to multi-sections and viceversa. Getting: "Item cannot be pasted into restricted containers."

This is by design (though you can tell me which things you were trying to copy to be sure). The multi section doesn't include all blocks. For example, I never want to see a hero carousel in tabs...

> 
> Will continue looking tomorrow if needed.

PR #583, opened 2026-02-09
536 concept animation

### LINKED ISSUE #536: Mayura "The Concept" modal animation

techmaven:
Change animation of the concept card modal from fade in/out to slide up/down.

Fix #536 

Test URLs:
- After: https://536-concept-animation--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/blocks/hero-heritage-cc/hero-heritage-cc.css
+++ b/blocks/hero-heritage-cc/hero-heritage-cc.css
@@ -304,7 +304,7 @@
   width: 100%;
   height: 100%;
   z-index: 3;
-  transition: opacity 0.35s ease-out, transform 0.35s ease-out;
+  transition: opacity 0.35s ease-out, transform 0.35s ease-out, visibility 0.35s ease-out;
   transform: scale(1);
 }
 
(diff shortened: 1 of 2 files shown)


roperev:
Ross, I think the abruptness of the animation is coming from the fact that all elements of ours just pop into and out of existence, rather then "sliding" up from underneath the invisible cover that's at the bottom of the button. If you watch very closely on the customer page (video attached), you'll see the whole "concept" div slides out from just below the "the concept" text, as if it were a "pocket" in the page, so to speak, that we're just pulling the div out of then sliding back into it. 

https://github.com/user-attachments/assets/86252fe1-3551-482a-a911-9612ff882a21

On your version, you're only doing the vertical movement and masking the existing in-page content. You'd either need to add a second clip path mask (probably to a different div, the "The concept" div perhaps?) that hides the bottom of it (always) and then when it moves up or moves down it will only show up in the main content area, or you could add some quick opacity faders, at least to the fade out, so when the content is dismissed, it slides down and fades out at the same time. 

https://github.com/user-attachments/assets/6034114f-430a-4b87-bbaa-575b3d73eac4

I think since this is the main hero, we do need just a bit more polish on this component.

helms-charity (approved):
Looks better with the fade now, approved.

PR #585, opened 2026-02-09
Add entrance animation for sections on viewport entry

### LINKED ISSUE #584: Implement a section loading animation style
Implement a section loading animation style that can be chosen by the author.

Already "Entrance Animation"  option has been added in _section.json for the authors to select : https://github.com/aemsites/idfc/issues/559

This is to perform the styling

Image

sthirumal:
Implemented an IntersectionObserver to add a visibility class to sections with the 'entrance-animation' class when they enter the viewport. Updated CSS for initial hidden state and visible state transitions.

Fix #584

Test URLs:
- After: https://584-secanim--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -787,7 +787,7 @@ export function decorateSections(parent, isDoc) {
 }
 
 /**
- * swapna: Observes .section.entrance-animation and adds .is-visible when section enters viewport.
+ * Observes .section.entrance-animation and adds .is-visible when section enters viewport.
  * @param {Element} container - Element to search for sections (e.g. main)
  */
 function initEntranceAnimationObserver(container) {
@@ -798,7 +798,7 @@ function initEntranceAnimationObserver(container) {
       entries.forEach((entry) => {
         if (entry.isIntersecting) {
           entry.target.classList.add('is-visible');
-          observer.unobserve(entry.target); /* swapna: stop observing once animated */
+          observer.unobserve(entry.target); /* stop observing once animated */
         }
       });
     },
@@ -868,7 +868,7 @@ export function decorateMain(main) {
   decorateIcons(main);
   buildAutoBlocks(main);
   decorateSections(main);
-  /* swapna: Init entrance-animation observer (animate when in viewport) */
+  /* Init entrance-animation observer (animate when in viewport) */
   initEntranceAnimationObserver(main);
   decorateBlocks(main);
   buildEmbedBlocks(main);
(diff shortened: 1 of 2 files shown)


sthirumal:
Following is the logic of this implementation :

Summary of what was added:
1. CSS in styles/styles.css
.section.entrance-animation – initial state: opacity: 0, transform: translateY(50px), transition: opacity 2s, transform 2s.
.section.entrance-animation.is-visible – visible state: opacity: 1, transform: translateY(0) so the section fades in and moves up.
2. JS in scripts/scripts.js
- initEntranceAnimationObserver(container) – uses IntersectionObserver to add .is-visible to .section.entrance-animation when it enters the viewport (with a small bottom rootMargin so it triggers just before fully in view).
- It’s run from decorateMain() right after decorateSections(main), so all such sections are observed on load.

Result for section (<div class="section entrance-animation ...">):
It starts at opacity: 0 and translateY(50px).
When it scrolls into view, it gets .is-visible and animates to opacity: 1 and translateY(0) over 2 seconds.


Performance assessment due to this implementation :


Viewport detection
Low
IntersectionObserver is the right, efficient API.

Animation
Low
Only opacity and transform; compositor can handle them.

Callback work
Low
Single classList.add, no layout reads.

Memory
Low
One observer, few sections; unobserve after animation keeps it minimal.

Initial load
Low
One querySelectorAll and observe() per section at startup.

scripts/scripts.js
```diff
@@ -786,6 +786,27 @@ export function decorateSections(parent, isDoc) {
   });
 }
 
+/**
+ * swapna: Observes .section.entrance-animation and adds .is-visible when section enters viewport.
```
helms-charity: you can remove your Swapna comments. We can see what you changed because the code is green :-P

helms-charity (approved):

helms-charity (changes requested):
There's a big hit to performance on your branch, please check that your code isn't interfering?
Update: I don't think it has to do with your PR, so I will approve.
image

helms-charity (approved):

sthirumal:
@helms-charity We tried 2 different ways to resolve conflict and merge and the page is getting broken. Need feedback on the code that has been added.

techmaven:
@sthirumal appears the conflict is due to your code insertions overlapping with insertions by Charity's multi section changes. I recommend using either the web editor or command line editor to open the conflicting file and shift your changes in scripts.js so there is no overlap.

techmaven:
Also have you tried asking Cursor to resolve the conflict?

PR #587, opened 2026-02-09
Add function to rewrite relative links to production origin in script…

### LINKED ISSUE #556: Convert local links to www.
We should not have any links to ww2. See them by searching for "ww2" on this page:
https://ww2.idfcfirst.bank.in/credit-card/metal-credit-card/mayura

No, there is no Edge Delivery Services setting that makes relative links on ww2.customer-domain.com automatically point to www.customer-domain.com instead. Relative links are always resolved against the current origin (host) the browser is on.

So if all our relative links in question are in the header, we can look for these and change them with JS (add to "decorate(block)")

```
const PROD_ORIGIN = 'https://www.customer-domain.com';

function makeProdUrl(href) {
  // ignore absolute URLs and non-http schemes
  if (!href || /^[a-zA-Z][a-zA-Z0-9+.-]*:/.test(href)) {
    return href;
  }
  // relative URL → prepend prod origin
  return new URL(href, PROD_ORIGIN).toString();
}

export default function decorate(block) {
...
  block.querySelectorAll('a[href]').forEach((a) => {
    a.href = makeProdUrl(a.getAttribute('href'));
  });
...
}
```

also we don't want to grab anchor links, so need something like `if (href.startsWith('#')) return href;` in there.

### Another item to consider
Maybe another ticket --
To prevent /content/idfcfirstbank/en links when we link to internal pages, we can add something in scripts.js. I found this in another project:

export function getPagePath(url) {
  return url ? new URL(url).pathname.replace(/\/content\/(dir1|dir2|dir3)|\.page|\.html/g, '') : '';
}

zhummerz:
…s.js

Fix #556 

Test URLs:
- After: https://556-convert-local-links--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -24,6 +24,14 @@ const MEDIA_QUERIES = {
   desktop: window.matchMedia('(min-width: 990px)'),
 };
 
+const PROD_ORIGIN = 'https://www.idfcfirst.bank.in';
+
+function makeProdUrl(href) {
+  if (!href || href.startsWith('#')) return href;
+  if (/^[a-zA-Z][a-zA-Z0-9+.-]*:/.test(href)) return href;
+  return new URL(href, PROD_ORIGIN).toString();
+}
+
 /**
  * Open PDF links and my.idfcfirst links in a new tab
  * @param {Element} a the anchor element to process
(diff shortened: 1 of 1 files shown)


sthirumal:
Code wise its looking good.

When tried to test below linking using link gopher extension and not seeing ww2. And based on my understanding on main branch the ww2 list should be listed.

https://main--idfc--aemsites.aem.page/credit-card/rupay-credit-card

helms-charity (approved):
Thanks for all the testing by you and Swapna.

PR #588, opened 2026-02-09
Testimonial Cards Redux

### LINKED ISSUE #133: Testimonial cards enhancements

bunting-adbe:
Break out of testimonial cards into their own block. The look and feel should be the same, but the authoring experience has been modified per the issue.  I have left the old block authored and the old code alone for comparison and validation. 

Fix #133 

Test URLs:
- After: https://133-testimonialEnhancement--idfc--aemsites.aem.page/credit-card/rupay-credit-card
- Library Before: https://main--idfc--aemsites.aem.page/tools/sidekick/library.html?plugin=blocks&path=/library/cards-testimonial&index=0
-Library After: https://133-testimonialenhancement--idfc--aemsites.aem.page/tools/sidekick/library.html?plugin=blocks&path=/library/cards-testimonial&index=0 

Desktop (Old above, New below):
image

Mobile (Old above, New below):
image

### DIFF
--- a/blocks/cards-testimonial/cards-testimonial.js
+++ b/blocks/cards-testimonial/cards-testimonial.js
@@ -1,0 +1,722 @@
+import {
+  createOptimizedPicture, loadScript, loadCSS, getMetadata,
+} from '../../scripts/aem.js';
+import { moveInstrumentation } from '../../scripts/scripts.js';
+
+function sanitizeText(text) {
+  if (!text) return '';
+  return (
+    String(text)
+      // eslint-disable-next-line no-control-regex
+      .replace(/[\x00-\x1F\x7F-\x9F]/g, '')
+      .replace(/\s+/g, ' ')
+      .replace(/[\u2018\u2019]/g, "'")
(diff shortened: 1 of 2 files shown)


blocks/cards-testimonial/cards-testimonial.js
```diff
@@ -0,0 +1,722 @@
...
+          }
+        });
+      });
+    });
+  };
+  updateStarIcons();
+  swiper.on('slideChange', updateStarIcons);
+  swiper.on('slideChangeTransitionEnd', updateStarIcons);
+
+  if (isMayuraTemplate) {
```
helms-charity: you can remove the Mayura template specifics.

helms-charity (changes requested):

blocks/cards-testimonial/cards-testimonial.js
bunting-adbe: removed

helms-charity (approved):
I think we can probably externalize some of the swiper functions in a new consolidation ticket, but this looks good for this ticket.

PR #589, opened 2026-02-09
Refactor table block to use section background handling code

### LINKED ISSUE #373: Update table styles
1. The Table's background color logic should be updated to use the same as the Section.

See the functions called `handleBackgroundImages` and `handleBackgroundImages` from scripts.js. Those are currently specific to sections, but with help from Cursor you should be able to apply to the table block too. Take note of what the JSON names are for those fields for color and images, the ones for Table might not be the exact same name as Section.

Image

https://main--idfc--aemsites.aem.page/library/table

roperev:
Normalized background color handling and improve image management. 
Added normalizeBackgroundColor function and updated handleBackground and handleBackgroundImages usage for better consistency.

Fix #373 

# Table vs Section background — shared logic

Table block now uses the same `handleBackground` and `handleBackgroundImages` from `scripts.js` as Section. Letter-case and JSON field names differ between the two; both are normalized before calling the shared functions.

---

## 1. JSON / schema field names (letter-case)

| Purpose | Section (metadata keys) | Table (block schema) |
|---------|-------------------------|----------------------|
| Background color | `backgroundcolor` | `backgroundColor` |
| Desktop image | `sectionBackgroundImage` | `image` |
| Mobile image | `sectionBackgroundImageMobile` | `imageMobile` |

---

## 2. How Table maps into the shared functions

- **Color:** Table reads `backgroundColor` from the block, normalizes bare hex (e.g. `fff` → `#fff`), then calls `handleBackground({ text: normalizedColor }, imageWrapper)` — same as Section's `handleBackground(metadata.backgroundcolor, section)`.
- **Images:** Table passes desktop/mobile URLs (from `image` / `imageMobile`) to `handleBackgroundImages(desktopUrl, mobileUrl, imageWrapper)`. If only mobile is set, Table passes `desktopUrl = mobileImageUrl` so the function still receives a desktop URL.

---

## 3. DOM after update (Section vs Table)

Both Section and Table now produce the same background structure when images are used:

**Section (e.g. .section)**

```html
<div class="section has-bg-images">
  <picture class="bg-images">
    <!-- source: mobile, tablet, desktop -->
    <img class="section-img" ...>
  </picture>
  <!-- ... block content ... -->
</div>
```

**Table (e.g. .table-background-image)**

```html
<div class="table-background-image has-bg-images">
  <picture class="bg-images">
    <!-- source: same breakpoints as Section -->
    <img class="section-img" ...>
  </picture>
</div>
<table>...</table>
```

---

## 4. Differences summary

| Aspect | Before (Table only) | After (shared with Section) |
|--------|---------------------|-----------------------------|
| **Color** | Custom hex/gradient logic, no `color-token` or `setColorScheme` | `handleBackground`: color-token support, `setColorScheme`, same as Section |
| **Images** | `createResponsivePicture` in table.js (different breakpoints) | `handleBackgroundImages`: same `createSource` + MEDIA_QUERIES as Section |
| **Classes on wrapper** | `.table-background-image` only | `.table-background-image` + `.has-bg-images`; inner `.bg-images`, `.section-img` |

---

*Reference only. Run the site and use a Table block with background color and/or images to verify (e.g. [library/table](https://main--idfc--aemsites.aem.page/library/table)).*

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/table
- After: https://373-tablestyles--idfc--aemsites.aem.page/library/table

### DIFF
--- a/blocks/table/table.js
+++ b/blocks/table/table.js
@@ -9,6 +9,7 @@ import {
   handleBackgroundImages,
   handleBackground,
   normalizeBackgroundColor,
+  getColorScheme,
 } from '../../scripts/scripts.js';
 
 /**
@@ -207,6 +208,14 @@ export default async function decorate(block) {
 
     if (imageAlt) imageWrapper.dataset.imageAlt = imageAlt;
     block.append(imageWrapper);
+
+    // setColorScheme(imageWrapper) only applies to imageWrapper's direct children (the picture).
+    // The table is a sibling, not a child, so apply scheme to block and table so content gets it.
+    const scheme = getColorScheme(imageWrapper);
+    if (scheme) {
+      block.classList.add(scheme);
+      table.classList.add(scheme);
+    }
   }
 
   if (thead) table.append(thead);
--- a/blocks/table/table.js
+++ b/blocks/table/table.js
@@ -8,6 +8,7 @@ import {
   moveInstrumentation,
   handleBackgroundImages,
   handleBackground,
+  normalizeBackgroundColor,
 } from '../../scripts/scripts.js';
 
 /**
@@ -20,19 +21,6 @@ function extractImageUrl(element) {
   if (element.tagName === 'IMG') return element.src;
   if (element.tagName === 'PICTURE') return element.querySelector('img')?.src || null;
   return null;
-}
-
-/**
- * Normalizes color for shared handleBackground (Table: backgroundColor; Section: backgroundcolor).
- * Ensures bare hex (e.g. "fff") gets "#" prefix so CSS is valid.
- * @param {string} color - Raw color from table metadata
- * @returns {string} - Color string suitable for handleBackground
- */
-function normalizeBackgroundColor(color) {
-  if (!color || typeof color !== 'string') return color;
-  const trimmed = color.trim();
-  if (trimmed.match(/^[0-9a-fA-F]{3,6}$/)) return `#${trimmed}`;
-  return trimmed;
 }
 
 /**
--- a/blocks/table/table.js
+++ b/blocks/table/table.js
@@ -4,7 +4,11 @@
  * https://www.hlx.live/developer/block-collection/table
  */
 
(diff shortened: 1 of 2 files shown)


roperev:
I tested this as throughly as possible, and I couldn't see any breaking changes to CSS or currently authored backgrounds/tables anywhere. Should be ready for review.

blocks/table/table.js
```diff
@@ -19,32 +23,16 @@ function extractImageUrl(element) {
...
-  img.alt = alt;
-  img.loading = 'lazy';
-  if (defaultImgUrl) img.src = defaultImgUrl;
-
-  picture.appendChild(img);
-  return picture;
+function normalizeBackgroundColor(color) {
+  if (!color || typeof color !== 'string') return color;
+  const trimmed = color.trim();
+  if (trimmed.match(/^[0-9a-fA-F]{3,6}$/)) return `#${trimmed}`;
```
helms-charity: I remember leaving this logic out of the sections, I think because it interfered with the "dark vs. light" luminance determination, but I didn't try very hard. Can you check if putting this function in Scripts.js (See parseColor, https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L445-L461) and removing the `if(!section) ` limitation will still allow the convert to RGB to work for both table and section? 

(I'm thinking ahead here, that we might as well allow the table to assign itself as dark-scheme or light-scheme independently of the section, but not for this ticket).

helms-charity (changes requested):
Thanks,

blocks/table/table.js
roperev: I've updated this now, moved the function and also handled the update to parseColor to allow tables or sections.

helms-charity (approved):
looking good

roperev:
Fixed remaining color-scheme detection and class application logic. Seems to be working properly now, see: https://373-tablestyles--idfc--aemsites.aem.page/library/table

image

PR #602, opened 2026-02-11
Add color tokens and hardcoded colors audit documentation

### LINKED ISSUE #580: Add all colors to tokens
See if Cursor can be trusted to:

- find all color values (hex or rgb) in Styles.css and for CSS files that are under /blocks ONLY. (not under /templates, /tools, etc).

- create new CSS tokens where the names start with --color, or for gradients, that they start with --gradient. The rest of the name Cursor can make it up as long as they are unique.

- do NOT tell Cursor to find and replace the colors with the new token names yet. You want to check the work. Do the color names make sense at all? Are some of the colors so close that we might be able to combine any?

Then make a PR for just this (without replacing anything) so I can have a look.

sthirumal:
**Update 2026-05-11T08:21:05-04:00 - Charity keeping this here as documentation, have been implementing parts of this slowly.**

---

css token and css color value of style.css and /blocks css and the css token suggestion.

Fix #580

Test URLs:
- After: https://580-tokcolor--idfc--aemsites.aem.page/

Please refer to COLOR_TOKENS_HARDCODED_AUDIT.md file

### DIFF
--- a/COLOR_TOKENS_AND_HARDCODED_AUDIT.md
+++ b/COLOR_TOKENS_AND_HARDCODED_AUDIT.md
@@ -1,0 +1,553 @@
+# Color tokens and hardcoded colors audit
+
+Scope: **styles/styles.css** and **CSS under /blocks only** (excluding /templates, /tools, etc.).
+
+---
+
+## Part 1: All CSS tokens and their color values (hex or rgb)
+
+Defined in `styles/styles.css` `:root` only. Gradients list the hex/rgb values used inside them.
+
+### Color tokens (single value)
+
+| Token | Color value |
+|-------|-------------|
+| `--color-red-300` | `#d13d3f` |
+| `--color-red-400` | `#9c1d26` |
+| `--color-red-500` | `#902a2b` |
+| `--color-red-600` | `#801c1e` |
+| `--color-red-800` | `#6e0a0f` |
+| `--color-blue-200` | `#007bff` |
+| `--color-blue-800` | `#2d3679` |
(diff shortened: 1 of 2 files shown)


sthirumal:
Please refer to COLOR_TOKENS_HARDCODED_AUDIT.md


## Summary of this file 

- **Part 1:** All color-related tokens are defined in `styles/styles.css` in `:root`: 21 single-value color tokens, 7 semantic aliases, and 16 gradient tokens (including `--modal-backdrop-gradient`). The tables above list each token and the hex/rgb values they use.
- **Part 2:** Every line listed above uses at least one hex or rgb color that is **not** coming from a token defined in `styles/styles.css`. Replacing these with new or existing tokens in `styles.css` would centralize colors and make them easier to change.
- **Part 3:** Suggested new `--color-*` and `--gradient-*` tokens derived from Part 2. Add the ones you need to `styles.css` and replace hardcoded values in the listed locations.

sthirumal:
@helms-charity  Added part 4 

Part 4: Part 3 colors are grouped by perceptual similarity (Delta E ≤ 1). Within each group, one canonical value and one token can replace all listed values and locations without visible change. Use the summary table to pick one token per group and reduce token count while keeping appearance consistent.

Please review it.

Let me know if you would like me to increase deltaE value more than 1.

sthirumal:
**Part 5:** Same Part 4 groups with an additional **Locations (file: line)** column listing the exact file path and line numbers where each suggested token (or the color it replaces) is used. Use Part 5 for find-and-replace when applying tokens.

PR #607, opened 2026-02-11
Refactor header height management for app banner visibility

### LINKED ISSUE #592: remove space for category nav if not used
### problem
On pages where there is no category nav, an awkward space is still left in the header. 
https://main--idfc--aemsites.aem.page/library/cards-testimonial

Image

I believe this is a regression from when I added the appBanner logic in loadEager at https://github.com/aemsites/idfc/blob/main/scripts/scripts.js#L1054-L1058 . I changed the default header height at styles.css:157 but did not create a new style override for desktop width.

Image

### To fix
I believe adding the header height here will fix the problem, but please check that there's no CLS increase on the Rupay and Mayura pages (which do have category navs on them).

Image

zhummerz:
Updated header height logic to conditionally adjust based on app banner state. When the app banner is closed, the header now uses only the nav-height to prevent spacing on mobile. Added corresponding CSS rules to ensure proper layout across different screen sizes.

Fix #592 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/cards-testimonial
- After: https://592-remove-space-for-category--idfc--aemsites.aem.page/library/cards-testimonial

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -944,10 +944,15 @@ async function loadEager(doc) {
 
   const getAppBanner = sessionStorage.getItem('getAppBanner');
   const header = doc.querySelector('header');
-  // When app banner was closed previously, use nav-height only (avoids space on mobile)
-  if (getAppBanner && header && !MEDIA_QUERIES.desktop.matches) {
-    header.style.height = 'var(--nav-height)';
-    document.body.classList.add('app-banner-closed');
+  if (header && !MEDIA_QUERIES.desktop.matches) {
+    if (getAppBanner) {
+      // Banner was closed previously: use nav-height only (avoids gap on mobile)
+      header.style.height = 'var(--nav-height)';
+      document.body.classList.add('app-banner-closed');
+    } else {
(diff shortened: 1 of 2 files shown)


zhummerz:
This was added to scripts.js to help with CLS on mobile. We get a shift score ding when the app CTA comes up when first loaded. Once you click to close it, it shifts up. This change in the js is to help with the session and only have the app CTA open once during the session:
 
if (header && !MEDIA_QUERIES.desktop.matches) {
    if (getAppBanner) {
      // Banner was closed previously: use nav-height only (avoids gap on mobile)
      header.style.height = 'var(--nav-height)';
      document.body.classList.add('app-banner-closed');
    } else {
      // Reserve space for banner before it loads (minimizes CLS when banner injects in loadLazy)
      document.body.classList.add('expect-app-banner');
    }

sthirumal (approved):
Looking good. Approved.

helms-charity (changes requested):
hang on, this is a lot more code than I was thinking....

helms-charity (approved):
Ok I had seen a bad mobile score for the Rupay page with your branch, but I ran it again and it seems fine. Approved.

PR #617, opened 2026-02-12
Card Date fix

### LINKED ISSUE #606: Testimonial card UE fixes
- we can remove the image and imageAlt fields since they aren't using them.
- that will bring us down to 6 model fields. For further help on how to get to 4, I've opened a thread at https://adobedx.slack.com/archives/C05QU7MMRNF/p1770815329589709 .
- What can we find out about that date? I assume the date was entered as yyyy-mm-dd, but I didn't expect it to change when it got to the JCR. [Check to see if there a documented way we should be entering dates](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/implementing/developing/universal-editor/field-types) here? I guess as long as the date is showing as expected in the LD+JSON (yyyy-mm-dd) and on the card, we're fine...

Image

Image

Image

bunting-adbe:
Added additional date properties to keep the date only and try and keep the time out of the JCR. 

Fix #606 

Test URLs:
- After: https://606-testimonialUE--idfc--aemsites.aem.page/

### DIFF
--- a/blocks/cards-testimonial/_cards-testimonial.json
+++ b/blocks/cards-testimonial/_cards-testimonial.json
@@ -123,6 +123,8 @@
           "valueType": "date",
           "description": "Review date. Shown on page in readable format.",
           "placeholder": "YYYY-MM-DD",
+          "displayFormat": "YYYY-MM-DD",
+          "valueFormat": "YYYY-MM-DD",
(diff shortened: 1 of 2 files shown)


techmaven (approved):
Looks good.

PR #634, opened 2026-02-13
Auto playback and loop video on UPI tabs

### LINKED ISSUE #626: tabs-upi-link video picker
http://localhost:3000/library/upitabs - library page. This block is only on the Rupay page.

There are layers to this bug.
1. Right now if you don't include an absolute link to an mp4, then no content shows on the first tab when you enter the page

Image

2. The reason the "temporary" text component is there in the first place is because I forgot we couldn't get a video file to work using the DAM picker.

Image

## Solution
First, fix the code so that the display of the first tab content isn't dependent upon having a link to a mp4 in that text component. If we didn't want a video on the phone, then we still want the tabbed content to show.

Then, research how we should be trying to choose an mp4 file. Perhaps we need to add something about the mime-type in that reference component? Check https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/implementing/developing/universal-editor/field-types .

bunting-adbe:
Make video from asset picker work on the page and loop playback.
This change is still in testing, but the video is not available on the page. Console errors report a 404 on the file. Fluffyjaws suggested the change to the paths.json to ensure the video was reachable by EDS. The JS changes are to remove the unnecessary video function that we removed with the UE updates. The scripts update was due to the size metadata not matching properly. 

Fix #626 

Test URLs:
- After: https://626-tabsUPI-videoPicker--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/tabs-upi-link/tabs-upi-link.js
+++ b/blocks/tabs-upi-link/tabs-upi-link.js
@@ -44,6 +44,32 @@ function getVideoOnlyPath(tabPanel) {
 const tabVideoMap = new Map();
 
 /**
+ * Extracts phone video URL from a block-level div (phone group: picture + video reference).
+ * Removes the video link so it does not display. Supports reference (a[href]) and legacy temp URL.
+ * @param {HTMLElement} phoneGroupDiv - Direct child of block that has picture and video link
+ * @returns {string|null} Absolute video URL or null
+ */
+function getPhoneVideoUrlFromPhoneGroup(phoneGroupDiv) {
+  const link = phoneGroupDiv.querySelector('a[href*=".mp4"]');
+  if (link && link.href) {
+    const url = link.href;
+    const p = link.closest('p');
+    if (p) p.remove();
+    else link.remove();
+    return url;
+  }
+  const p = phoneGroupDiv.querySelector('p');
+  if (p) {
+    const text = p.textContent.trim();
+    if (text.endsWith('.mp4') && (text.startsWith('http://') || text.startsWith('https://'))) {
+      p.remove();
+      return text;
+    }
+  }
+  return null;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


blocks/tabs-upi-link/tabs-upi-link.js
```diff
@@ -85,13 +85,27 @@ export default async function decorate(block) {
...
+    : null;
+
+  // Single source for picture: prefer image-only cell (old), else phone group (new)
+  const pictureSourceDiv = imageDiv || phoneGroupDiv;
+
+  // Extract video URL from phone group (new model) and remove the link from DOM
+  if (phoneGroupDiv) {
+    const videoUrl = getPhoneVideoUrlFromPhoneGroup(phoneGroupDiv);
+    if (videoUrl) block.dataset.phoneVideoUrl = videoUrl;
+  }
```
roperev, line 108: Per the comment here, it's supposed to remove the link from the DOM, but it's just adding it to the block dataset properties. Does it need to also remove it or is it ok this way? Either the comment or the function should be updated accordingly.

roperev (changes requested):
Aside from the commented part, everything looks good.

blocks/tabs-upi-link/tabs-upi-link.js
bunting-adbe: I have several other JS changes I am not ready to merge, but these were already committed. I can make the edit but don't want to commit them until I see if any of this will even work with the update to the paths.

bunting-adbe: Comment updated to `// Extract video URL from phone group`
Will be added to next commit.

roperev (approved):
All good

PR #635, opened 2026-02-14
UPI Video Picker Enhancement

### LINKED ISSUE #626: tabs-upi-link video picker

bunting-adbe:
With the previous change (PR 634) we added a new path to the paths.json which allowed the video to be playable, but it is only working in the UE (this can be seen now without the below code). The code in this PR specifically cleans up some logic and allows for the use of the fallback , from the repo, video to be used on the rendered page. This is necessary due to a limitation of EDS to make MP4's available at the edge. 
Working with Fluffyjaws it seems this is known and the workarounds it was aware of did not work for this case (https://experienceleague.adobe.com/en/docs/experience-cloud-kcs/kbarticles/ka-27519#token_type=bearer&expires_in=28799997). 
This change makes the video work, but in a state that is not controlled by authors, as the video needs to be available in the code repo.  

Fix #626 

Test URLs:
- After: https://626-tabsUPI-videoPicker--idfc--aemsites.aem.page/credit-card/rupay-credit-card

### DIFF
--- a/blocks/tabs-upi-link/tabs-upi-link.js
+++ b/blocks/tabs-upi-link/tabs-upi-link.js
@@ -3,45 +3,65 @@ import { toClassName } from '../../scripts/aem.js';
 import { moveInstrumentation } from '../../scripts/scripts.js';
 
 /**
- * Creates a video element for the phone animation (loop, muted, autoplay).
- * Uses <source> + setAttribute for reliable playback across browsers.
- * @param {string} videoSrc - The video source URL
- * @returns {HTMLVideoElement} The video element
+ * Returns a fallback video URL for DAM paths. When the primary URL is /content/dam/...,
+ * EDS/preview often returns 404; the fallback points to the same filename under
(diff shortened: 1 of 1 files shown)


helms-charity (approved):

helms-charity (approved):
Thanks for the research.

PR #645, opened 2026-02-17
Add Subtitle option to UE for Section-Title

### LINKED ISSUE #642: Section-title: Add additional subtitle options

Image

bunting-adbe:
Initial UE updates for the sub-title of the section-title block. I also generated some updates to the styling to account for this update, but need the UE changes to be added to fully test and validate all styling works as expected. 

Fix #642 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/section-title
- After: https://642-sectionTitle-options--idfc--aemsites.aem.page/drafts/section-title

### DIFF
--- a/blocks/section-title/_section-title.json
+++ b/blocks/section-title/_section-title.json
@@ -23,102 +23,97 @@
       "fields": [
         {
           "component": "text",
-          "name": "title",
-          "label": "Title",
+          "name": "title_text",
+          "label": "Title Text",
           "valueType": "string"
         },
         {
           "component": "select",
-          "name": "titleType",
+          "name": "title_type",
           "label": "Title Type",
           "value": "h2",
           "valueType": "string",
           "options": [
-            {
-              "name": "h1",
-              "value": "h1"
-            },
-            {
-              "name": "h2",
-              "value": "h2"
-            },
-            {
-              "name": "h3",
-              "value": "h3"
-            },
-            {
-              "name": "h4",
-              "value": "h4"
-            },
-            {
-              "name": "h5",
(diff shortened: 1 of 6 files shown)


blocks/section-title/_section-title.json
```diff
@@ -23,107 +23,102 @@
       "fields": [
         {
           "component": "text",
-          "name": "title",
```
helms-charity, line 26: keep title

blocks/section-title/_section-title.json
```diff
@@ -23,107 +23,102 @@
...
-          "label": "Title",
+          "name": "title_text",
+          "label": "Title Text",
           "valueType": "string"
         },
         {
           "component": "select",
-          "name": "titleType",
```
helms-charity, line 32: keep titleType

blocks/section-title/_section-title.json
```diff
@@ -23,107 +23,102 @@
...
+            { "name": "h3", "value": "h3" },
+            { "name": "h4", "value": "h4" },
+            { "name": "h5", "value": "h5" },
+            { "name": "h6", "value": "h6" },
+            { "name": "p", "value": "p" }
           ]
         },
         {
           "component": "select",
-          "name": "textSize",
```
helms-charity, line 69: change to title_size

helms-charity (changes requested):

helms-charity (approved):
I'm approving so you don't get stuck, but you should end up with this I think:
title, titleType and title_size
subtitle, subtitleType, subtitle_size

PR #651, opened 2026-02-18
Reorder button group decoration in main function for improved functio…

### LINKED ISSUE #639: Bug: button groups break the UE
See /library/buttons
When the superscript for button groups is authored, the text component is split up, and the text that was grouped can go missing.

### Test that this is truly the problem first.
You shouldn't trust what I was seeing in just the /library/buttons page, you should create a draft page from scratch with the text editor and add all the tests of the buttons with and then without the superscript grouping treatment to validate there is a problem, and that the problem only exists when there are button groups.

Image

  
Image

zhummerz:
…nality

Fix #639 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/buttons
- After: https://639-button-group-break--idfc--aemsites.aem.page/drafts/buttons

### DIFF
--- a/scripts/scripts.js
+++ b/scripts/scripts.js
@@ -974,12 +974,12 @@ function decorateLinkedPictures(main) {
 export function decorateMain(main) {
   // hopefully forward compatible button decoration
   decorateButtons(main);
-  decorateButtonGroups(main);
   decorateIcons(main);
   decorateSections(main); /* must be before buildAutoBlocks */
   buildAutoBlocks(main);
   initEntranceAnimationObserver(main);
   decorateBlocks(main);
+  decorateButtonGroups(main);
   buildEmbedBlocks(main);
   decorateLinkedPictures(main);
 }


zhummerz:
The Problem: decorateButtonGroups runs at line 977 in decorateMain, before decorateSections (line 979) and decorateBlocks (line 982). This creates a div.button-group at the section level. Then:

1. groupChildren (called by decorateSections) treats any DIV with a className as a "block". So div.button-group gets put into a block-content wrapper, and the surrounding text gets split into separate default-content wrappers. This is why the text gets split up and goes missing.
2. decorateBlocks matches div.button-group via the selector div.section > div > div:not(.section), so decorateBlock is called on it — adding data-block-name="button-group" and data-block-status="initialized". That's why data-block* attributes appear.

The Fix: Move decorateButtonGroups(main) to run after decorateBlocks(main). At that point, the content is already grouped into wrappers, and the button-group div will be created safely inside a default-content wrapper — never seen by groupChildren or decorateBlocks.

helms-charity (approved):
I can't see the fix when using ref=639-button-group-break, but I no longer see the "block" reference in the DOM for these button groups, so I'm approving.

PR #663, opened 2026-02-20
UE allow cards-icon in sections

### LINKED ISSUE #657: Move some card types into new block
I'm thinking these stay as "cards", or maybe we call them "cards-icon"? (keeping the word 'cards' first on purpose so all cards show up together in the UE list).

Image

and we split the ones that do carousel-ing and modal-ing into a new cards block (probably "cards-carousel"?)

Image

Image

Image

helms-charity:
I forgot to allow the new cards-icon under sections in the UE.

Fix #657 

Test URLs:
- After: https://657-cardsicon--idfc--aemsites.aem.page/

### DIFF
--- a/component-filters.json
+++ b/component-filters.json
@@ -35,7 +35,8 @@
       "overview",
       "section-title",
       "hotspot",
-      "cards-testimonial"
+      "cards-testimonial",
+      "cards-icon"
     ]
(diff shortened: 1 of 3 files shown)


roperev (approved):
classic

PR #666, opened 2026-02-20
subtitle works

### LINKED ISSUE #642: Section-title: Add additional subtitle options

bunting-adbe:
Subtitle option works. Header tag and size configs also. 

Fix #642 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/section-title-lives
- After: https://642-sectiontitle-options--idfc--aemsites.aem.page/drafts/section-title-lives

Main:
image

Branch:
image

### DIFF
--- a/blocks/section-title/section-title.css
+++ b/blocks/section-title/section-title.css
@@ -6,76 +6,38 @@
   padding: 0;
 }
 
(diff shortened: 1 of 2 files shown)


techmaven (approved):
Looks good

PR #669, opened 2026-02-20
Logic for using Icons

### LINKED ISSUE #642: Section-title: Add additional subtitle options

bunting-adbe:
Updated logic to account for use of Icons in titles

Fix #642 

Test URLs:
- Before: https://main--idfc--aemsites.aem.page/drafts/section-title-lives
- After: https://642-sectiontitle-options--idfc--aemsites.aem.page/drafts/section-title-lives

Main:
image

Branch:
image

### DIFF
--- a/blocks/section-title/section-title.js
+++ b/blocks/section-title/section-title.js
@@ -72,15 +72,22 @@ function getHeadingFromCell(cell) {
       text: (heading.textContent ?? '').trim(),
       tag: heading.tagName.toLowerCase(),
       id: heading.id ?? '',
+      innerHTML: heading.innerHTML?.trim() || '',
     };
   }
-  return { text: cellText(cell), tag: 'h2', id: '' };
+  return {
+    text: cellText(cell),
+    tag: 'h2',
+    id: '',
+    innerHTML: '',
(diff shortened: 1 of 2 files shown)


blocks/section-title/section-title.js
```diff
@@ -113,11 +120,13 @@ export default function decorate(block) {
   }
 
   if (rows.length >= 2) titleSizeClass = normalizeSize(cellText(rows[1])) || titleSizeClass;
+  let subtitleHtml = '';
   if (rows.length >= 3) {
     const sub = getHeadingFromCell(rows[2]);
     if (hasValue(sub.text)) {
       subtitleText = sub.text;
       subtitleTag = sub.tag;
+      subtitleHtml = sub.innerHTML || '';
```
helms-charity: I don't think this is right. How was it working before? we shouldn't allow HTML to be entered.

bunting-adbe: Cursor help get that sorted and use the icon function. I thought I had told it that before, but we made it there eventually.

techmaven (approved):
Looks good.

PR #672, opened 2026-02-21
Hotspot.js: reduce complexity and add DOMpurify

### LINKED ISSUE #671: hotspot: security and maintainability fixes
- functions should not be nested more than 4 levels deep
- use DOMpurify on innerHTML, etc to prevent XSS

helms-charity:
- DOMpurify related functions in scripts.js were already merged per #673 but are repeated here. They are needed to sanitize document write methods.
- applied the new sanitizeHTML to the places we were using innerHTML in hotspot.js 
So content from DOM/dataset/contentMap is sanitized both when it’s first read/stored and again when it’s written to innerHTML, which addresses the aikido.dev finding.

- Reduced nested functions for less complexity and better maintainability. 

Summary of what was refactored with regard to nested functions:

1. runAfterDoubleRAF(fn)
Single helper that runs fn after two requestAnimationFrame calls, so the “double RAF then run” logic lives in one place.

2. schedulePadding
A small function that runs runAfterDoubleRAF(calculateInitialPadding), used both as the image load listener and in the “already loaded” path.

3. Removed waitForImageAndCalculate
Replaced with a direct if/else: if the image isn’t loaded yet, register schedulePadding on load; otherwise call runAfterDoubleRAF(calculateInitialPadding) immediately.

4. fadeTransition (was ~L300)
Introduced makeFadeInEndHandler() so the transition-end logic is no longer nested inside startFadeIn.
startFadeIn now only adds the class and attaches the handler returned by makeFadeInEndHandler().
Kept the double-RAF call as a one-liner: requestAnimationFrame(() => requestAnimationFrame(startFadeIn)).

5. Go-back link handling (was ~L748 and ~L947)
Added a shared handleGoBackLinkClick(evt, sectionData, container, reattachListeners) that:
Reads href and targetId. If no targetId: expands the hero and returns.
Otherwise loads content from sectionData.contentMap, runs fadeTransition, then calls the passed-in reattachListeners and showInitialPanel.
Both the initial decorate() setup and reattachHotspotListeners() now do:
link.addEventListener('click', (evt) => handleGoBackLinkClick(evt, sectionData, container, reattachHotspotListeners)).

Fix #671 

Test URLs:
- After: https://671-hotspotsecurity--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

Testing:
There should be no difference in how The Concept animation works.
In your IDE, you should see this (when only showing errors is checked)
image


helms-charity:
Cursor's notes on reducing complexity from 22 to 15:

Summary of the refactor that brought cognitive complexity under the limit:
1. New helpers (used by decorate and/or reattachHotspotListeners)
getStoredOrLiveContent(sectionData, targetId) – Returns HTML from the section’s contentMap or from the live block in the DOM.
applyClickPadding(container, hotspot) – Updates tooltip padding so the content moves toward the clicked hotspot during fade-out.
navigateToHotspotContent(container, targetId, hotspot, reattachListeners) – Runs the full link flow: resolve content, add to transitioningContainers, apply padding, run fadeTransition, then reattach listeners and show the panel.
2. Helpers used only in decorate
parseBlockMetadata(rows) – Reads the first three rows and returns { imageElement, blockId, hotspotText }, replacing the previous chain of if (rows.length > N) blocks.
handleStandardHotspotClick(...) – Handles a non-link hotspot click: deactivate others, toggle panel visibility, set panel HTML, attach go-back handlers.
updateSectionMinHeight(sectionWrapper, block, sectionData, blockHeight) – Updates sectionData.maxHeight and the section/block-content minHeight when the block is taller.
3. decorate()
Uses parseBlockMetadata(rows) for metadata.
Link hotspots: click handler only calls navigateToHotspotContent(...).
Standard hotspots: click handler only calls handleStandardHotspotClick(...).
Section height logic is delegated to updateSectionMinHeight(...).
4. reattachHotspotListeners()
Link branch now only calls navigateToHotspotContent(container, targetId, hotspot, reattachHotspotListeners) instead of inlining the same logic.
Cognitive complexity is now within the allowed 15, and there are no remaining linter errors in this file.

iustinp (approved):
LGTM

PR #676, opened 2026-02-21
Scripts.js maintenance: sanitizeHTML and reduce function complexities

### LINKED ISSUE #675: scripts.js XSS and high sonarQube fixes
use DOMpurify:
User controlled data in methods like 'innerHTML', 'outerHTML' or 'document.write' is an anti-pattern that can lead to XSS vulnerabilities.

After updating the sonarLint and secure-code plugins:
Image

helms-charity:
The biggest change was that `handleSectionMetadata `was reduced from a complexity of 41 to under 15 by moving logic into five helpers:
- applySpecialSectionMetadata(metadata, section) – style, backgroundcolor, grid, gap, spacing, containerwidth, height/heightmobile.
- applyDataAttributesFromMetadata(metadata, section) – writes non-special keys to section.dataset and sets section.id when present.
- applySectionBackgroundImages(metadata, section) – section desktop/mobile background images.
- applyDecorationImages(metadata, section) – decoration (doodle) images and reverse flag.
- applyBlockContentMetadata(metadata, section) – block-content background, images, object-fit, object-position.
- addressed CWE-915 / OWASP A01 – Prototype pollution. Replaced the plain object with a Map so block IDs are stored with blockIds.get(name) / blockIds.set(name, value) instead of blockIds[name]. No prototype is involved, so there’s no prototype pollution risk.

Also added the section metadata key lists as Set()s since apparently this has been a thing since ES6.

Fix #675 

Test URLs:
- After: https://675-scriptsjsmain--idfc--aemsites.aem.page/

- After: https://675-scriptsjsmain--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

Testing: There should not be any functional difference.

image
After: your IDE should not have errors anymore.
image


amarghioali (approved):
LGTM

PR #706, opened 2026-02-24
Updated all the prev JS via browser-security

### LINKED ISSUE #705: Second scan innerHTML findings
Apparently I missed some things on the first pass. Also Aikido is still flagging all the instances of sanitizeHTML I added, which are safe because they are going through DomPurify, but I don't know how to tell this tool that.

### Using document.write related methods
TL;DR
User controlled data in methods like 'innerHTML', 'outerHTML' or 'document.write' is an anti-pattern that can lead to XSS vulnerabilities.

How do I fix it?
Make sure you do not use user input when you are using these methods and if you do, be sure to sanitize the input.

Line 481 in cards-testimonial.js
reviewWrapper.innerHTML = reviewHtml;

Line 96 in icons.js
cardIcon.innerHTML = icon.svg;

Line 1537 in header.js
exploreLink.innerHTML = `Explore ${titleText} <span class="arrow">→</span>`;

### Using window.location
TL;DR
Using window.location.href or similar redirection methods with unvalidated user input can lead to cross-site scripting (XSS). If an attacker controls part of the URL, they may inject malicious scripts that execute in the victim’s browser, potentially allowing session hijacking, data theft, or other unauthorized actions. Proper validation and encoding of user-controlled values are required before using them in redirects or client-side navigation.

How do I fix it?
Avoid using window.location.href or similar redirection methods with unvalidated user input. Only allow redirects to trusted, predefined destinations or validate the input against a strict allowlist. When possible, use fixed URLs or server-side redirection logic, and ensure any user-controlled values are properly validated and encoded before use.

Line 593 in header.js
window.location.href = `https://www.idfcfirst.bank.in/search?skey=${encodeURIComponent(query)}`;

Line 834 in header.js
buttonLi.addEventListener('click', () => { window.location.href = link.href; });

helms-charity:
I learned of another plugin that I should have used all along, and with that I needed to rename "sanitizeHTML" to DOMPurify.sanitize() so it would be deemed a "trusted sanitizer".
In some cases, we only needed to use "textContent" instead of "innerHTML" so I didn't need to sanitize at all.

So there should be no missing functionality, just better security with our new rules in eslintrc.js.

Fix #705 

Test URLs:
- After: https://705-innerhtml--idfc--aemsites.aem.page/

- After: https://705-innerhtml--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

New tickets were created for the remaining blocks that didn't get the first pass treatment.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
   root: true,
-  plugins: ['secure-coding', 'sonarjs'],
+  plugins: ['secure-coding', 'browser-security', 'sonarjs'],
   extends: [
     'airbnb-base',
     'plugin:json/recommended',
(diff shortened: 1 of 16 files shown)


bunting-adbe (approved):
I don't see anything of concern. Looks good.

PR #720, opened 2026-02-25
Enhance security for editor-support.js

### LINKED ISSUE #716: editor.js security
Image

roperev:
- Category navigation handling enhanced by limiting block loading to 100 items
- Added new functions for injecting and decorating category-nav sections
- Improved error logging for main element updates.

Fix #716 

Test URLs:
- After: https://716-editorsecurity--idfc--aemsites.aem.page/

This file is what is loaded in UE window. I loaded it in my `ref=branch` view, but I don't have a great way of validating everything. Everything within normal page editing type functions seemed to behave as expected, nothing seemed to be broken or throwing errors.

### DIFF
--- a/scripts/editor-support.js
+++ b/scripts/editor-support.js
@@ -96,9 +96,11 @@ async function reloadCategoryNav(main) {
     console.error('[Category Nav Editor] Failed to load CSS:', error);
   }
 
-  // Load the category-nav blocks to trigger their decoration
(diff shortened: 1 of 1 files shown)


helms-charity (approved):
I think this looks good. I don't know that we can really test it without merging into main, so I'll do that and we can always revise if needed.

PR #721, opened 2026-02-25
cards.js reduce complexity and other linting

### LINKED ISSUE #709: cards.js security and complexity
Image

helms-charity:
Besides breaking out complex functions into smaller utility functions, I found this fix interesting:

Problem:
⚠️ CWE-915 OWASP:A01-Broken CVSS:9.8 | Object injection/Prototype pollution (incl. model/tool outputs) | MEDIUM [SOC2,PCI-DSS,ISO27001]
   Fix: Use Map or property whitelisting | https://portswigger.net/web-security/prototype-pollution

**Fix applied:**
Line 296: cardBodies[0] → cardBodies.item(0)
Line 297: cardBodies[cardBodies.length - 1] → cardBodies.item(cardBodies.length - 1)
Why this works
The linter (CWE-915) flags bracket property access like obj[key] because it can’t tell that key is safe, so it warns about possible prototype pollution.
NodeList.item(index) is the standard, index-based API for NodeList. It returns the same element as cardBodies[index] but uses a method call instead of bracket notation, so the linter no longer reports it.
Behavior is unchanged: item(i) and [i] are equivalent for valid indices on a NodeList.
.at(-1) can break in older environments where NodeList doesn’t support at(); item(length - 1) is supported everywhere.
So the warning at line 297 is addressed without changing behavior or relying on .at(-1).

Fix #709 

Test URLs:
- After: https://709-cardsjsxss--idfc--aemsites.aem.page/

- After: https://709-cardsjsxss--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

There should be no functional changes to the cards.

### DIFF
--- a/blocks/cards/cards.js
+++ b/blocks/cards/cards.js
@@ -1,7 +1,7 @@
 import {
   createOptimizedPicture, loadScript, loadCSS, toCamelCase,
 } from '../../scripts/aem.js';
-import { moveInstrumentation } from '../../scripts/scripts.js';
+import { moveInstrumentation, DOMPURIFY } from '../../scripts/scripts.js';
 import { createModal } from '../modal/modal.js';
 
 /**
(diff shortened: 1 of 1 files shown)


amarghioali (approved):
LGTM

PR #723, opened 2026-02-25
category-nav.js security and complexity

### LINKED ISSUE #710: category-nav.js security and complexity
Image

helms-charity:
- The arrows didn't need to have innerHTML
- the regex didn't need to be a bracketed set for just 1 character'
- complexity was reduced in two places:
1. parseSectionWithCardsBlock (was ~687)
getTextElementsOutsideBlocks(section, cached) – Resolves “use cache or query” and the loop that filters text elements outside blocks.
createIconElementFromSpan(iconSpan) – Builds the icon <span> and  and returns { iconElement, iconName }.
The main function now does: get text elements → parse first (icon vs category name) → dropdown title → find cards block → convert cards → return. Fewer branches and no extra work.
2. decorate (was ~796)
getSectionsWithCardsData(allCategoryNavBlocks) – Contains the “if we have nav blocks, get container, iterate siblings, check cards/icon” logic and returns the { section, textElements } array.
decorate now calls this helper and then continues with the rest of the flow. Early returns are unchanged; the previous nested if (allCategoryNavBlocks.length > 0) { if (container) { forEach(...) } } is moved into the helper.

Fix #710 

Test URLs:
- After: https://710-categorynavxss--idfc--aemsites.aem.page/

Testing: the category nav functionality should be unchanged.

### DIFF
--- a/blocks/category-nav/category-nav.js
+++ b/blocks/category-nav/category-nav.js
@@ -181,7 +181,7 @@ export function parseCategoryNavBlock(block) {
 
   return {
     title: categoryName,
-    id: categoryName.toLowerCase().replace(/\s+/g, '-').replace(/[&]/g, ''),
+    id: categoryName.toLowerCase().replace(/\s+/g, '-').replace(/&/g, ''),
     eyebrowTitle,
     linkText,
     linkUrl,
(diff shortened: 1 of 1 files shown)


amarghioali (approved):
LGTM

PR #725, opened 2026-02-25
tabs-upi-link.js: Sanitize HTML and reduce complexity

### LINKED ISSUE #714: tabs-upi-link security and complexity
Image

helms-charity:
**Summary of the CWE-79 fix that keeps tab icons working:**
1. Imports
Import ensureDOMPurify and DOMPURIFY from ../../scripts/scripts.js.
2. Load DOMPurify before using it
At the start of decorate(block), after the early return, add await ensureDOMPurify() so the tab label HTML is only set after DOMPurify is available.
3. Sanitize tab label HTML
Build the label HTML: const tabLabelHtml = `\<p>${tabNameElement.innerHTML}</p>\;`
Set the button content from the sanitized result only:
button.innerHTML = window.DOMPurify.sanitize(tabLabelHtml, DOMPURIFY);
DOMPurify is used with the project’s DOMPURIFY config (USE_PROFILES: { html: true }), which keeps safe HTML (e.g. `, <picture>`) and strips script and other dangerous content. Tab labels can still include icons/images, but script injection is removed.

**Summary of the cognitive complexity refactoring:**
1. getPhoneVideoUrlFromPhoneGroup (was at line 67)
removeMp4LinkFromDom(link) – Removes the link (or its parent p) from the DOM.
getVideoUrlFromMp4Link(link) – Gets the URL from an a[href*=".mp4"], normalizes it with URL, removes the link, and returns the URL (or null).
getVideoUrlFromLegacyParagraph(phoneGroupDiv) – Handles the legacy case where the URL is in a paragraph’s text (localhost/https + .mp4).
getPhoneVideoUrlFromPhoneGroup – Now just: try link, then try legacy paragraph; complexity is reduced.
2. decorate (was at line 100)
getBlockStructure(children) – Returns { titleDiv, imageDiv, phoneGroupDiv, pictureSourceDiv } so the “find title/image/phone/picture” logic lives in one place.
buildTabButtons(block, tabPanels, tablist, tabNames) – Builds the tablist and tab buttons (the previous processableTabs.forEach body).
buildImageAndMobilePanels(pictureSourceDiv, tabPanels, tabsWrapper) – Builds the image block and mobile panel layout (the previous if (pictureSourceDiv) block with nested if (picture) and tabPanels.forEach).
attachVideosToBlock(block) – Handles fallback URL and attaches video elements to .phone-animation-video-container.

Fix #714 

Test URLs:
- After: https://714-tabsupixss--idfc--aemsites.aem.page/

Testing: there should not be any functional changes.

image

### DIFF
--- a/blocks/tabs-upi-link/tabs-upi-link.js
+++ b/blocks/tabs-upi-link/tabs-upi-link.js
@@ -1,6 +1,6 @@
 // eslint-disable-next-line import/no-unresolved
 import { toClassName } from '../../scripts/aem.js';
-import { moveInstrumentation } from '../../scripts/scripts.js';
+import { ensureDOMPurify, moveInstrumentation, DOMPURIFY } from '../../scripts/scripts.js';
 
 /**
  * Returns a fallback video URL for DAM paths. When the primary URL is /content/dam/...,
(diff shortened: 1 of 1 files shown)


bunting-adbe (approved):
approved

PR #749, opened 2026-02-27
new cards-icon block

### LINKED ISSUE #746: cards-icon block
now that all the refactoring and linting is over, I want to finish separating out the simpler cards into a new block.

Image

helms-charity:
Went back to the OOTB cards code to for these since they are much simpler that our other cards.
- They are now using generic names for the style classes:  image-left, image-top, and image-left-arrow 
- I created a new library page with the new block, but after merging I will need to author the Mayura page and then remove the redundant code that I copied from Cards.

image

Fix #746
Test URLs:
- Before: https://main--idfc--aemsites.aem.page/library/cards-icon
- After: https://746-cardsicon--idfc--aemsites.aem.page/library/cards-icon

### DIFF
--- a/blocks/cards-icon/cards-icon.css
+++ b/blocks/cards-icon/cards-icon.css
@@ -2,33 +2,20 @@
    CARDS-ICON BLOCK
    Simpler icon-card block for three layout variants:
      - image-left        : icon on left, no border/background
-     - important-documents : icon on top, grid of white-bg cards
-     - related-search    : icon on left with right-arrow indicator
+     - image-top          : icon on top, grid of white-bg cards
+     - image-left-arrow   : icon on left with right-arrow indicator
    Mobile-first.
    ================================================================= */
-
-/* --- Shared utilities --- */
-
-.cards-icon .sr-only {
-  position: absolute;
-  width: 1px;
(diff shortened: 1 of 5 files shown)


roperev:
Looks like the gap between the new "important-cards" (image-left) is a bit too big, should be 40px, currently is 66px. Also the font size seems just a little bit to small on the new version. 

image

blocks/cards-icon/cards-icon.css
```diff
@@ -300,105 +268,82 @@
...
+    justify-content: space-between;
     margin-left: auto;
     margin-right: auto;
   }
 
-  .cards-icon.important-documents .cards-icon-card {
+  .cards-icon.image-top .cards-icon-card {
     width: 175px;
     height: 175px;
     padding: var(--spacing-s) 26px;
```
roperev, line 333: This might be adding your "extra" spacing? 40px gap + 26px spacer = 66px which is what's displaying currently: 

Image

roperev (changes requested):
Good, just some spacing and font size adjustments needed.

roperev (approved):
All good.

PR #772, opened 2026-03-10
testing bg

helms-charity:
With this change, I am seeing much better mobile scores.

Before:
image

After:
image

Summary of changes for LCP and resource load delay:

1. Preload at the start of `<head>`
The preload link is now inserted with document.head.insertBefore(preloadLink, document.head.firstChild) instead of appendChild. The browser sees the LCP image preload earlier, which can reduce load delay.

2. Use a real `` for the LCP image
The background is no longer applied only via background-image on the section. The first picture is moved into a dedicated background layer and kept in the DOM as a real `` with:

fetchpriority="high" – asks the browser to prioritize this image
loading="eager" – avoids lazy-loading so it starts loading immediately
PageSpeed and other tools treat visible `` elements as LCP candidates and prioritize them; CSS background images are often discovered and prioritized later. Keeping the hero image as a`n ` in a full-bleed layer improves the chance it’s treated as LCP and loads earlier.

3. New background layer and CSS
A div with class hero-heritage-cc-bg-layer is inserted as the first child of the section.
The first picture (and its wrapper) is moved into this layer instead of being removed.
CSS makes the layer full-viewport with position: fixed; inset: 0; z-index: -1 and the image object-fit: cover so the look matches the previous fixed, cover background.

Test URLs:
- After: https://769-svgtexture--idfc--aemsites.aem.page/credit-card/metal-credit-card/mayura

- Before: https://main--idfc--aemsites.aem.page/credit-card/metal-credit-card/ashva
- After: https://769-svgtexture--idfc--aemsites.aem.page/credit-card/metal-credit-card/ashva

image

### DIFF
--- a/blocks/hero-heritage-cc/hero-heritage-cc.css
+++ b/blocks/hero-heritage-cc/hero-heritage-cc.css
@@ -9,6 +9,31 @@
   overflow: hidden;
 }
 
+/* LCP background layer: real img for better load priority than CSS background-image */
+.hero-heritage-cc-bg-layer {
+  position: fixed;
+  inset: 0;
+  z-index: -1;
+  overflow: hidden;
+  max-width: 100vw !important;
+}
+
+.hero-heritage-cc-bg-layer picture,
+.hero-heritage-cc-bg-layer p {
+  position: absolute;
+  inset: 0;
+  margin: 0;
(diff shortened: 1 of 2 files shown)


roperev (approved):
Should be great

